PR #2, opened 2023-07-19
Add support for entire database namespaces

aldaliu:
For our usecase, we are doing migrations of entire DBs and would like to provide a DB name as a namespace and have the migration-verifier check all collections of the DB.

Here are the high level changes:
- Extracted a function to find all namespaces for a database from ListAllUserCollections in `list_namespaces.go`
- I've added some logic to the parsing of the --srcNamespace flag to detect database names in `check.go`
    - for each database found, use the new function to get all the collections in that DB and append it to the slice of srcNamespace
- added some util functions to help parse namespaces for DB names


sibelius:
any progress on this ?

PR #3, opened 2023-07-20
REP-2942 Verifier: Add user-defined query filtering to the initial check logic

tdq45gj:
Added `globalFilter` field to the Verifier for user-defined filter. The filter is applied when fetching documents.

### DIFF
--- a/internal/partitions/partition_test.go
+++ b/internal/partitions/partition_test.go
@@ -23,8 +23,8 @@ func (suite *UnitTestSuite) TestPartitionFindCmd() {
 				{"afterClusterTime", startAt},
 			}},
 			{"noCursorTimeout", true},
-			{"filter", expectedFilter},
 			{"hint", bson.D{{"_id", 1}}},
+			{"filter", expectedFilter},
 		}
 		actual := partition.FindCmd(suite.Logger(), startAt)
 		assertBSONEqual(suite.T(), expectedFind, actual)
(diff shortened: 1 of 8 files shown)


internal/verifier/migration_verifier.go
```diff
@@ -392,22 +397,27 @@ func (verifier *Verifier) getGenerationWhileLocked() (int, bool) {
 	return verifier.generation, verifier.lastGeneration
 }
 
+func (verifier *Verifier) maybeAppendGlobalFilter(predicates bson.A) bson.A {
```
huan-Mongo, line 400: The name of the function is a bit confusing, Other readers might think  `globalFilter` is changed in this function.

internal/verifier/migration_verifier.go
```diff
@@ -392,22 +397,27 @@ func (verifier *Verifier) getGenerationWhileLocked() (int, bool) {
...
-		}
+		andPredicates = append(andPredicates, bson.D{{"_id", bson.M{"$in": task.Ids}}})
+		andPredicates = verifier.maybeAppendGlobalFilter(andPredicates)
 		findOptions = bson.D{
-			bson.E{"filter", filter},
+			bson.E{"filter", bson.D{{"$and", andPredicates}}},
 		}
 	} else {
-		findOptions = task.QueryFilter.Partition.GetFindOptions(buildInfo)
+		findOptions = task.QueryFilter.Partition.GetFindOptions(buildInfo, verifier.maybeAppendGlobalFilter(andPredicates))
```
huan-Mongo, line 420: Is `verifier.maybeAppendGlobalFilter(andPredicates)` same as `verifier.globalFilter` here?

huan-Mongo (changes requested):
Couple of questions on the implementation.

internal/verifier/migration_verifier.go
tdq45gj: No, `maybeAppendGlobalFilter` should return predicates (`bson.A`), instead of the global filter (`bson.D`). I changed the name of the method to `maybeAppendGlobalFilterToPredicates` for clarification.

huan-Mongo (approved):
LGTM!

FGasper (changes requested):

internal/partitions/partition_test.go
```diff
@@ -23,8 +23,8 @@ func (suite *UnitTestSuite) TestPartitionFindCmd() {
 				{"afterClusterTime", startAt},
 			}},
 			{"noCursorTimeout", true},
-			{"filter", expectedFilter},
 			{"hint", bson.D{{"_id", 1}}},
+			{"filter", expectedFilter},
```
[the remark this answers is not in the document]
tdq45gj: I reordered the find options in the `GetFindOptions` method for code quality reason, so that the `filter` field is appended at the end for both cases instead of separately in each block of an `if` statement.

internal/verifier/check.go
```diff
@@ -122,18 +122,20 @@ func (verifier *Verifier) CheckWorker(ctx context.Context) error {
...
 		verifier.logger.Info().Msg("Verifier already checking the collections")
 		return nil
 	}
 	verifier.running = true
+	verifier.globalFilter = filter
 	verifier.mux.Unlock()
 	defer func() {
 		verifier.mux.Lock()
 		verifier.running = false
+		verifier.globalFilter = nil
```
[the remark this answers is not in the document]
tdq45gj: It's unset so that the filter is only used in the context of the current running of `CheckDriver` method. The next call of `CheckDriver` might run with a different `globalFilter`.

FGasper (approved):
LGTM

tdq45gj:
evergreen merge
Added globalFilter field to the Verifier for user-defined filter. The filter is applied when fetching documents.

PR #4, opened 2023-08-02
REP-2943 / REP-2944: Add `filter` parameter to `/check` endpoint and enable query filtering 

tdq45gj:
The `/check` endpoint takes a JSON filter. The verifier parses the JSON filter to `map[string]any`. The verifier starts a change stream without query filtering. The verifier generates a recheck document for a change event of a filtered document, but the document should be filtered out during fetching and thus not compared by the verifier.

### DIFF
--- a/internal/partitions/partition.go
+++ b/internal/partitions/partition.go
@@ -190,16 +190,16 @@ func (p *Partition) GetFindOptions(buildInfo *bson.M, filterAndPredicates []bson
 func (p *Partition) filterWithNoTypeBracketing() bson.D {
 	// We use $expr to avoid type bracketing and allow comparison of different _id types,
 	// and $literal to avoid MQL injection from an _id's value.
-	return bson.D{{"$and", bson.A{
+	return bson.D{{"$and", []bson.D{
 		// All _id values >= lower bound.
-		bson.D{{"$expr", bson.D{
+		{{"$expr", bson.D{
 			{"$gte", bson.A{
 				"$_id",
 				bson.D{{"$literal", p.Key.Lower}},
 			}},
 		}}},
 		// All _id values <= upper bound.
-		bson.D{{"$expr", bson.D{
+		{{"$expr", bson.D{
 			{"$lte", bson.A{
 				"$_id",
 				bson.D{{"$literal", p.Upper}},
@@ -212,10 +212,10 @@ func (p *Partition) filterWithNoTypeBracketing() bson.D {
 // partition.  This filter will not properly handle mixed-type _ids -- if the upper and lower
 // bounds are of different types (except minkey/maxkey), nothing will be returned.
 func (p *Partition) filterWithTypeBracketing() bson.D {
-	return bson.D{{"$and", bson.A{
+	return bson.D{{"$and", []bson.D{
 		// All _id values >= lower bound.
-		bson.D{{"_id", bson.D{{"$gte", p.Key.Lower}}}},
+		{{"_id", bson.D{{"$gte", p.Key.Lower}}}},
 		// All _id values <= upper bound.
-		bson.D{{"_id", bson.D{{"$lte", p.Upper}}}},
+		{{"_id", bson.D{{"$lte", p.Upper}}}},
 	}}}
 }
--- a/internal/partitions/partition.go
+++ b/internal/partitions/partition.go
@@ -137,7 +137,7 @@ func (p *Partition) FindCmd(
 // (e.g. use the partitions on the source to read the destination for verification)
 // If the passed-in buildinfo indicates a mongodb version < 5.0, type bracketing is not used.
 // filterAndPredicates is a slice of filter criteria that's used to construct the "filter" field in the find option.
-func (p *Partition) GetFindOptions(buildInfo *bson.M, filterAndPredicates []bson.D) bson.D {
+func (p *Partition) GetFindOptions(buildInfo *bson.M, filterAndPredicates bson.A) bson.D {
 	if p == nil {
 		if len(filterAndPredicates) > 0 {
 			return bson.D{{"filter", bson.D{{"$and", filterAndPredicates}}}}
--- a/internal/verifier/migration_verifier.go
+++ b/internal/verifier/migration_verifier.go
@@ -645,7 +645,7 @@ func (verifier *Verifier) ProcessVerifyTask(workerNum int, task *VerificationTas
 					dataSizes = append(dataSizes, v.dataSize)
 				}
 				// Update ids of the failed task so that only ids from mismatches are reported.
-				// Ids of checked documents are discarded and hidden from report if they passed verification.
+				// Ids of matching documents are discarded and hidden from the mismatching documents report.
 				task.Ids = ids
 				err := verifier.InsertFailedCompareRecheckDocs(task.QueryFilter.Namespace, ids, dataSizes)
 				if err != nil {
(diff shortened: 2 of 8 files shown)


internal/verifier/migration_verifier.go
```diff
@@ -399,8 +399,10 @@ func (verifier *Verifier) getGenerationWhileLocked() (int, bool) {
 
 func (verifier *Verifier) maybeAppendGlobalFilterToPredicates(predicates []bson.D) []bson.D {
 	if verifier.globalFilter == nil {
+		verifier.logger.Debug().Str("filter", fmt.Sprintf("%v", verifier.globalFilter)).Msg("No filter to append")
```
mmcclimon, line 402: If we're in this branch, we know that verifier.globalFilter is nil, and so we also know that `fmt.Sprintf("%v")` is always going to return `<nil>`. I wonder if it would be more useful to not include this Str() call at all, and just have the message say "No filter to append; globalFilter is nil" or something.

internal/verifier/migration_verifier_test.go
```diff
@@ -1428,3 +1428,97 @@ func (suite *MultiDataVersionTestSuite) TestGenerationalRechecking() {
...
+	<-checkDoneChan
+	status = waitForTasks()
+	// There should be no failures, since the inserted document is not in the filter.
+	suite.Require().Equal(VerificationStatus{TotalTasks: 1, CompletedTasks: 1}, *status)
+
+	// Now insert in the source, this should come up next generation.
+	_, err = srcColl.InsertOne(ctx, bson.M{"_id": 4, "x": 44, "inFilter": true})
+	suite.Require().NoError(err)
+
+	// Tell check to start the next generation.
```
mmcclimon, line 1498: These running comments are really useful for following this code I don't know well; thanks!

mmcclimon (changes requested):
This is looking good! I left some small comments, and I think having a few more tests for the web server would be good before we merge it, but those should be straightforward to add. 

(I will also give you my usual note that I'm not super familiar with the verifier; I haven't read all of the supporting code carefully to make sure you're using all the APIs correctly, but the code you've written here is sensible and the fact that you've added tests makes me much more confident than I'd be without them, so thanks!)

FGasper (changes requested):

internal/verifier/migration_verifier.go
```diff
@@ -399,8 +399,10 @@ func (verifier *Verifier) getGenerationWhileLocked() (int, bool) {
 
 func (verifier *Verifier) maybeAppendGlobalFilterToPredicates(predicates []bson.D) []bson.D {
 	if verifier.globalFilter == nil {
```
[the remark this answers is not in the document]
tdq45gj: I think it's like a find or `$match` query `{}` in that case, so an empty slice is a valid filter and the intended behavior is to match every document.

internal/verifier/migration_verifier.go
```diff
@@ -399,8 +399,10 @@ func (verifier *Verifier) getGenerationWhileLocked() (int, bool) {
...
+		verifier.logger.Debug().Str("filter", fmt.Sprintf("%v", verifier.globalFilter)).Msg("No filter to append")
 		return predicates
 	}
+	verifier.logger.Debug().Str("filter", fmt.Sprintf("%v", verifier.globalFilter)).Msg("Appending filter to find query")
```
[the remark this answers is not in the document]
tdq45gj: `globalFilter` is a single `bson.D` unmarshaled from a JSON object. Although it can contain multiple fields, I don't think the user can put a slice of filters.

internal/verifier/migration_verifier.go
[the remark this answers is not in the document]
FGasper: Doh, you’re right … not sure what I was thinking there.

internal/verifier/web_server.go
```diff
@@ -177,15 +185,38 @@ func (server *WebServer) Run(ctx context.Context) error {
...
 		c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
 		return
 	}
 
-	server.Mapi.Check(context.Background())
+	var filter bson.D
+	if req.Filter != nil {
+		raw, err := bson.Marshal(req.Filter)
+		if err != nil {
+			err = errors.Wrap(err, "failed to marshal json to bson.Raw")
```
FGasper, line 206: Nit: This isn’t marshalling _JSON_ per se. The fact that the `map[string]any` came from JSON isn’t relevant here.
```suggestion
			err = errors.Wrap(err, "failed to marshal map to bson.Raw")
```

internal/verifier/web_server.go
```diff
@@ -177,15 +185,38 @@ func (server *WebServer) Run(ctx context.Context) error {
...
+			c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
```
FGasper, line 207: Wouldn’t an error here be an internal server error (HTTP 500)? “Bad Request” would indicate that there’s something the user could do to change the outcome, but if all we’re doing here is turning a map into BSON bytes, that seems like something the user can’t influence; i.e., it’s internal to the server.

internal/verifier/web_server.go
```diff
@@ -177,15 +185,38 @@ func (server *WebServer) Run(ctx context.Context) error {
...
+			return
+		}
+
+		err = bson.Unmarshal(raw, &filter)
+		if err != nil {
+			err = errors.Wrap(err, "failed to unmarshal bson.Raw filter to bson.D")
+			c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
```
FGasper, line 214: Same comment as above: this seems more like HTTP 500, where the error wouldn’t be returned directly to the client.

FGasper (changes requested):

internal/verifier/web_server.go
tdq45gj: Now that we're using `map[string]any` directly for `filter`, this error does not occur in the web server.

internal/verifier/web_server.go
tdq45gj: No longer relevant because we don't marshal map here.

internal/verifier/web_server.go
tdq45gj: Same here, this code is removed.

internal/partitions/partition.go
```diff
@@ -137,7 +137,7 @@ func (p *Partition) FindCmd(
 // (e.g. use the partitions on the source to read the destination for verification)
 // If the passed-in buildinfo indicates a mongodb version < 5.0, type bracketing is not used.
 // filterAndPredicates is a slice of filter criteria that's used to construct the "filter" field in the find option.
-func (p *Partition) GetFindOptions(buildInfo *bson.M, filterAndPredicates []bson.D) bson.D {
+func (p *Partition) GetFindOptions(buildInfo *bson.M, filterAndPredicates bson.A) bson.D {
```
FGasper, line 140: This change makes this interface less type-specific. A caller could submit `bson.A{nil, false, bson.D{{"1", 2}}}`, for example, which is better avoided.

Consider instead adding a third parameter, e.g.:
```
func (p *Partition) GetFindOptions(buildInfo *bson.M, verifierConditions []bson.D, userFilter map[string]any)
```
… then combining those inside the function as `bson.D{{"$and", bson.A{ verifierConditions, userFilter }}}`.

FGasper (approved):

PR #6, opened 2023-10-10
REP-3423 Migration-verifier should retry on Location28799 error while sampling documents

tdq45gj:
This PR adds `Location28799` error and make it retryable if running a pipeline with `$sample` stage.

### DIFF
--- a/internal/verifier/migration_verifier.go
+++ b/internal/verifier/migration_verifier.go
@@ -398,7 +398,7 @@ func (verifier *Verifier) getGenerationWhileLocked() (int, bool) {
 }
 
 func (verifier *Verifier) maybeAppendGlobalFilterToPredicates(predicates bson.A) bson.A {
-	if verifier.globalFilter == nil {
+	if len(verifier.globalFilter) == 0 {
 		verifier.logger.Debug().Msg("No filter to append; globalFilter is nil")
 		return predicates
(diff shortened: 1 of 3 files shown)


autarch (approved):
LGTM!

internal/verifier/migration_verifier.go
```diff
@@ -398,7 +398,7 @@ func (verifier *Verifier) getGenerationWhileLocked() (int, bool) {
 }
 
 func (verifier *Verifier) maybeAppendGlobalFilterToPredicates(predicates bson.A) bson.A {
-	if verifier.globalFilter == nil {
+	if len(verifier.globalFilter) == 0 {
```
FGasper, line 401: Why this change?

tdq45gj: This is to prevent an error like [REP-3390](https://jira.mongodb.org/browse/REP-3390). I don't think it's necessary here since we are only appending the filter to a slice of other predicates. But I think it makes slightly more sense to compare length with 0 than to check for a nil slice.

FGasper (approved):
LGTM. Thanks!

tdq45gj:
evergreen merge 
This PR adds `Location28799` error and make it retryable if running a pipeline with `$sample` stage.

PR #7, opened 2023-10-12
REP-3431 Create partitions using _ids of filtered documents 

tdq45gj:
This PR adds a filter to the partition logic so that partition bounds are created by randomly sampling filtered documents. Number of partitions is calculated by multiplying a ratio of filtered documents with the entire-collection num of partitions. Since sampling of partition bounds will use filtered documents, they will be more evenly distributed across partitions. A more reasonable number of partitions will also be created if only a small subset of documents out of a collection are to be included in the filter. If the partitioning-after-filtering approach takes too long (timeout error), it will fall back to partitioning without filtering.

### DIFF
--- a/internal/partitions/partitions.go
+++ b/internal/partitions/partitions.go
@@ -134,7 +134,37 @@ func PartitionCollectionWithSize(
 		partitionSizeInBytes = defaultPartitionSizeInBytes
 	}
 
-	return PartitionCollectionWithParameters(ctx, uuidEntry, &retryer, srcClient, replicatorList, defaultSampleRate, defaultSampleMinNumDocs, partitionSizeInBytes, subLogger, globalFilter)
+	partitions, docCount, byteCount, err := PartitionCollectionWithParameters(
+		ctx,
+		uuidEntry,
+		&retryer,
+		srcClient,
+		replicatorList,
+		defaultSampleRate,
+		defaultSampleMinNumDocs,
+		partitionSizeInBytes,
+		subLogger,
+		globalFilter,
+	)
+
+	// Handle timeout errors by partitioning without filtering.
+	if mongo.IsTimeout(err) {
+		subLogger.Debug().Err(err).Msgf("Timed out while partitioning with filter (%+v), continuing by partitioning without the filter.", globalFilter)
+		return PartitionCollectionWithParameters(
+			ctx,
+			uuidEntry,
+			&retryer,
+			srcClient,
+			replicatorList,
+			defaultSampleRate,
+			defaultSampleMinNumDocs,
+			partitionSizeInBytes,
+			subLogger,
+			nil,
+		)
+	}
+
+	return partitions, docCount, byteCount, err
 }
 
 // PartitionCollectionWithParameters is the implementation for
@@ -197,7 +227,18 @@ func PartitionCollectionWithParameters(
 	// appropriate number of partitions.
 	numPartitions := 1
 	if !isCapped {
-		numPartitions = getNumPartitions(collSizeInBytes, partitionSizeInBytes)
+		// By default, number of partitions is calculated without considering the ratio of filtered documents.
+		numPartitions = getNumPartitions(collSizeInBytes, partitionSizeInBytes, 1)
+
+		// If a filter is used for partitioning, number of partitions is calculated with the ratio of filtered documents.
+		if len(globalFilter) > 0 {
+			numFilteredDocs, filteredCntErr := GetDocumentCountAfterFiltering(ctx, subLogger, retryer, srcColl, uuidEntry.UUID, globalFilter)
+			if filteredCntErr == nil {
+				numPartitions = getNumPartitions(collSizeInBytes, partitionSizeInBytes, float64(numFilteredDocs)/float64(collDocCount))
+			} else {
+				return nil, 0, 0, filteredCntErr
+			}
+		}
 	}
 
 	// Prepend the lower bound and append the upper bound to any intermediate bounds.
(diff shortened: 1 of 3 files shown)


internal/partitions/partitions.go
```diff
@@ -360,17 +384,24 @@ func getOuterIDBound(ctx context.Context, subLogger *logger.Logger, retryer *ret
 	}
 
 	var docID interface{}
+
+	var pipeline []bson.D
+	if len(globalFilter) > 0 {
+		pipeline = append(pipeline, bson.D{{"$match", globalFilter}})
+	}
+	pipeline = append(pipeline, []bson.D{
+		bson.D{{"$sort", bson.D{{"_id", sortDirection}}}},
```
FGasper, line 393: Nit: You can omit the outer `bson.D` here since it’s implicit in the slice type.

internal/partitions/partitions.go
```diff
@@ -426,7 +468,25 @@ func getMidIDBounds(ctx context.Context, logger *logger.Logger, retryer *retry.R
 		numDocsToSample = collDocCount
 	}
 
-	logger.Info().Msgf("Sampling %d documents to make %d partitions for collection '%s.%s', UUID %s", numDocsToSample, numPartitions, srcDB.Name(), collName, collUUID.String())
+	var msg string
+	var pipeline []bson.D
```
FGasper, line 472: Nit: Consider the `mongo.Pipeline` type.

internal/partitions/partitions.go
```diff
@@ -360,17 +384,24 @@ func getOuterIDBound(ctx context.Context, subLogger *logger.Logger, retryer *ret
 	}
 
 	var docID interface{}
+
+	var pipeline []bson.D
```
FGasper, line 388: Nit: Consider the `mongo.Pipeline` type.

internal/partitions/partitions.go
```diff
@@ -426,7 +468,25 @@ func getMidIDBounds(ctx context.Context, logger *logger.Logger, retryer *retry.R
...
 	}
 
-	logger.Info().Msgf("Sampling %d documents to make %d partitions for collection '%s.%s', UUID %s", numDocsToSample, numPartitions, srcDB.Name(), collName, collUUID.String())
+	var msg string
+	var pipeline []bson.D
+	if len(globalFilter) > 0 {
+		pipeline = append(pipeline, bson.D{{"$match", globalFilter}})
+		msg = fmt.Sprintf("Sampling documents with filter (%v)", globalFilter)
+	} else {
+		msg = fmt.Sprintf("Sampling %d documents", numPartitions)
```
FGasper, line 477: Should this be `numDocsToSample`?

internal/partitions/partitions.go
```diff
@@ -426,7 +468,25 @@ func getMidIDBounds(ctx context.Context, logger *logger.Logger, retryer *retry.R
...
+	}
+
+	logger.Info().Msgf("%s to make %d partitions for collection '%s.%s', UUID %s", msg, srcDB.Name(), collName, collUUID.String())
```
FGasper, line 480: The arguments here seem mis-aligned with the format string … ?

FGasper (changes requested):
Looks good generally. I’ve labeled a few (very minor) nits, and a format-string issue.

internal/partitions/partitions.go
```diff
@@ -360,17 +494,24 @@ func getOuterIDBound(ctx context.Context, subLogger *logger.Logger, retryer *ret
 	}
 
 	var docID interface{}
+
+	var pipeline mongo.Pipeline
+	if len(globalFilter) > 0 {
+		pipeline = append(pipeline, bson.D{{"$match", globalFilter}})
+	}
+	pipeline = append(pipeline, []bson.D{
```
jtcovan, line 502: Cool! I forgot ... can unpack the slice

internal/partitions/partitions.go
```diff
@@ -329,13 +384,82 @@ func GetSizeAndDocumentCount(ctx context.Context, logger *logger.Logger, retryer
...
+	logger.Debug().Msgf("Collection %s.%s filtered document count: %d, filter: %+v",
+		srcDB.Name(), currCollName, value.Count, filter)
+
+	return value.Count, nil
+}
+
+// getNumPartitions returns the total number of partitions needed for the collection,
+// which is proportional to the percentage of filtered documents in the collection.
 //
 // The returned number is always 1 or greater, where 1 indicates that the collection
```
jtcovan, line 458: Is the returned number aways >= 1 even with the ratio? I am guessing yes since we can't have 0 partitions?

internal/verifier/migration_verifier_test.go
```diff
@@ -1476,10 +1489,12 @@ func (suite *MultiDataVersionTestSuite) TestVerifierWithFilter() {
 	// Wait for one generation to finish.
 	<-checkDoneChan
 	status := waitForTasks()
-	suite.Require().Equal(VerificationStatus{TotalTasks: 2, FailedTasks: 0, CompletedTasks: 2}, *status)
+	suite.Require().Greater(status.CompletedTasks, 1)
+	suite.Require().Greater(status.TotalTasks, 1)
```
jtcovan, line 1493: Why is it Greater now instead of Equal 2? I'm not very familiar with the verifier

jtcovan (approved):
LGTM, one question

internal/partitions/partitions.go
tdq45gj: Yea, the return value adds 1 to a non-negative int.

internal/verifier/migration_verifier_test.go
tdq45gj: I think it was 2 because there's 1 task to verify a partition and another task to verify the collection. Now that we're changing this test to create many partitions, I think it's hard to test the exact number of tasks / partitions in this high-level test.

FGasper (approved):
LGTM. I left a couple minor suggestions; no need to review if you want to accept or reject them.

Thanks!

PR #8, opened 2023-10-19
REP-2682 Fix failure report with no shown failures.

FGasper:
It’s been common with the verifier for generation 0’s report to indicate “Mismatches found” but to show 0 documents that actually do mismatch. This changeset fixes that; specifically:

- UpdateVerificationTask() now persists the task’s Ids.
- ProcessVerifyTask() now populates Ids and FailedDocs as reportDocumentMismatches() expects.

Additionally, some comments are added to aid future maintenance, a few “unhelpful” iteration variables are renamed, and the leftover “Type” column header in the mismatched-documents table is removed.

### DIFF
--- a/internal/verifier/migration_verifier.go
+++ b/internal/verifier/migration_verifier.go
@@ -619,7 +619,7 @@ func (verifier *Verifier) compareOneDocument(srcClientDoc, dstClientDoc bson.Raw
 func (verifier *Verifier) ProcessVerifyTask(workerNum int, task *VerificationTask) {
 	verifier.logger.Debug().Msgf("[Worker %d] Processing verify task", workerNum)
 
-	mismatches, docsCount, bytesCount, err := verifier.FetchAndCompareDocuments(task)
+	problems, docsCount, bytesCount, err := verifier.FetchAndCompareDocuments(task)
 
 	if err != nil {
 		task.Status = verificationTaskFailed
@@ -628,7 +628,7 @@ func (verifier *Verifier) ProcessVerifyTask(workerNum int, task *VerificationTas
 		task.SourceDocumentCount = docsCount
 		task.SourceByteCount = bytesCount
 
-		if len(mismatches) == 0 {
+		if len(problems) == 0 {
 			task.Status = verificationTaskCompleted
 		} else {
 			task.Status = verificationTaskFailed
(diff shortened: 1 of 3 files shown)


internal/verifier/verification_task.go
```diff
@@ -59,11 +59,19 @@ type VerificationTask struct {
 	Type       verificationTaskType   `bson:"type"`
 	Status     verificationTaskStatus `bson:"status"`
 	Generation int                    `bson:"generation"`
-	Ids        []interface{}          `bson:"_ids"`
+
+	// For failed tasks, this stores the document IDs missing on
+	// one cluster or the other.
+	Ids []interface{} `bson:"_ids"`
```
autarch, line 65: nit: Maybe rename this to `MissingIDs`? Also note that Go style is to use all for abbreviations like "ID", as opposed to "Id".

internal/verifier/migration_verifier.go
```diff
@@ -638,16 +638,35 @@ func (verifier *Verifier) ProcessVerifyTask(workerNum int, task *VerificationTas
 					workerNum, task.PrimaryKey)
 			} else {
 				verifier.logger.Debug().Msgf("[Worker %d] Verification Task %+v failed, may pass next generation", workerNum, task.PrimaryKey)
-				var ids []interface{}
+
+				var mismatches []VerificationResult
+				var missingIds []interface{}
```
autarch, line 643: nit: "IDs".

internal/verifier/verification_task.go
```diff
@@ -59,11 +59,19 @@ type VerificationTask struct {
...
+
 	// Deprecated: VerificationTask ID field is ignored by the verifier.
-	ID          int                  `bson:"id"`
-	FailedDocs  []VerificationResult `bson:"failed_docs,omitempty"`
-	QueryFilter QueryFilter          `bson:"query_filter"          json:"query_filter"`
+	ID int `bson:"id"`
+
+	// For failed tasks, this stores details on documents that exist on
+	// both clusters but don’t match.
+	FailedDocs []VerificationResult `bson:"failed_docs,omitempty"`
```
autarch, line 72: Would it make sense to rename this to "MismatchedDocs"?

autarch (approved):
LGTM % some small nits. The new code makes sense to me though I don't 100% understand why the old code had the bug it did. No need for another review.

tdq45gj (approved):
LGTM

internal/verifier/verification_task.go
FGasper: Unfortunately this gets used for other purposes in other event types. I’ll rename `Ids` as you suggest.

internal/verifier/migration_verifier.go
FGasper: capitalize? That seems a bit funky.

internal/verifier/verification_task.go
FGasper: Possibly, but the verifier kind of overloads the distinction between “failed” and “mismatched”. I’m not sure that this doesn’t get alternately used in both cases. :-P

PR #10, opened 2024-03-19
REP-4197: fix collectionUUID not supported error for 5.0 versions

huan-Mongo:
In v5.0, the error "collectionUUID is not supported on a mongos" in the collstats command is not retried because the error is a "FailedToParse" error. I added the retry condition for the collectionUUID error in v5.0.

### DIFF
--- a/internal/retry/retry.go
+++ b/internal/retry/retry.go
@@ -181,12 +181,18 @@ func (r *Retryer) runRetryLoop(
 			continue
 		}
 
+		isPreV50CollectionUUIDNotSupportErr := func(err error) bool {
+			return util.IsFailedToParseError(err) && util.HasServerErrorMessage(err, "collectionUUID")
+		}
+		isV50CollectionUUIDNotSupportErr := func(err error) bool {
+			return util.HasServerErrorMessage(err, "collectionUUID is not supported")
+		}
+
 		// If this is the first time we've come across a failure to parse
 		// collection UUID, try again with the UUID elided (if the caller used
 		// RequestWithUUID).
 		if r.retryOnUUIDNotSupported && !r.aggregateDisallowsUUIDs &&
-			(util.IsFailedToParseError(err) && util.HasServerErrorMessage(err, "collectionUUID")) || // This error is returned in pre 5.0 server
-			util.HasServerErrorMessage(err, "collectionUUID is not supported") { // The error is returned in the v5.0 server
+			(isPreV50CollectionUUIDNotSupportErr(err) || isV50CollectionUUIDNotSupportErr(err)) {
 			logger.Debug().Msg("Server does not support UUIDs in 'aggregate'. Will retry without UUID.")
 			ri.attemptNumber++
 			r.aggregateDisallowsUUIDs = true
--- a/internal/retry/retry.go
+++ b/internal/retry/retry.go
@@ -184,7 +184,9 @@ func (r *Retryer) runRetryLoop(
 		// If this is the first time we've come across a failure to parse
 		// collection UUID, try again with the UUID elided (if the caller used
 		// RequestWithUUID).
-		if r.retryOnUUIDNotSupported && !r.aggregateDisallowsUUIDs && util.HasServerErrorMessage(err, "collectionUUID") {
+		if r.retryOnUUIDNotSupported && !r.aggregateDisallowsUUIDs &&
+			(util.IsFailedToParseError(err) && util.HasServerErrorMessage(err, "collectionUUID")) || // This error is returned in pre 5.0 server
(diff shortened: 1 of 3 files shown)


internal/retry/retry.go
```diff
@@ -184,8 +184,11 @@ func (r *Retryer) runRetryLoop(
 		// If this is the first time we've come across a failure to parse
 		// collection UUID, try again with the UUID elided (if the caller used
 		// RequestWithUUID).
-		if r.retryOnUUIDNotSupported && !r.aggregateDisallowsUUIDs && util.IsFailedToParseError(err) && util.HasServerErrorMessage(err, "collectionUUID") {
-			logger.Debug().Msg("Server version (< 5.0) does not support UUIDs in 'aggregate'. Will retry without UUID.")
+		if r.retryOnUUIDNotSupported && !r.aggregateDisallowsUUIDs &&
+			(util.IsFailedToParseError(err) && util.HasServerErrorMessage(err, "collectionUUID")) || // This error is returned in pre 5.0 server
```
tdq45gj, line 188: Can we split the logic into 2 variables, for example, `preV50collUUIDNotSupported` and `v50collUUIDNotSupported`, and add them to the debug log?

tdq45gj (changes requested):
I left a small suggestion. Thank you for the fix!

internal/retry/retry.go
huan-Mongo: Make sense. Done!

internal/retry/retryer_test.go
```diff
@@ -284,3 +284,31 @@ func (suite *UnitTestSuite) TestRetryerWithEmptyCollectionName() {
...
+			return nil
+		}
+	}
+
+	r := retryer.SetRetryOnUUIDNotSupported()
+	r.aggregateDisallowsUUIDs = false
+	_, err := r.RunForUUIDAndTransientErrors(suite.Context(), suite.Logger(), "bar", f)
+	// The aggregateDisallowsUUIDs will be set to True in the retry
+	suite.True(r.aggregateDisallowsUUIDs)
+	suite.Nil(err)
```
FGasper, line 312:
```suggestion
	suite.NoError(err)
```

internal/retry/retry.go
```diff
@@ -181,11 +181,20 @@ func (r *Retryer) runRetryLoop(
 			continue
 		}
 
+		isPreV50CollectionUUIDNotSupportErr := func(err error) bool {
+			return util.IsFailedToParseError(err) && util.HasServerErrorMessage(err, "collectionUUID")
+		}
+		isV50CollectionUUIDNotSupportErr := func(err error) bool {
+			return util.HasServerErrorMessage(err, "collectionUUID is not supported")
+		}
```
FGasper, line 189: Why define these as runtime anonymous subs rather than normal functions? Then they could be tested directly.

internal/retry/retryer_test.go
```diff
@@ -284,3 +284,31 @@ func (suite *UnitTestSuite) TestRetryerWithEmptyCollectionName() {
...
+// Test fix for REP-4197
+func (suite *UnitTestSuite) TestRetryerWithUUIDNotSupportedError() {
+	retryer := New(0)
+
+	attemptNumber := 0
+	cmdErr := mongo.CommandError{
+		Message: "(Location4928902) collectionUUID is not supported on a mongos",
+		Name:    "FailedToParseError",
+		Code:    9,
+	}
```
FGasper, line 297: Is it possible to test against both v5 and pre-v5 variants of the error?

FGasper (changes requested):
This seems fine in general. I‘ve pointed out some minor concerns.

internal/retry/retry.go
huan-Mongo: I don't think these functions worth testing and now they are not called anywhere else. We can move them to normal functions if they are shared later.

FGasper: FWIW, I respectfully disagree that they should be defined inline unless they need to be called elsewhere.

Making these unexported, named functions will minimize bloat in this function and prevent inadvertent closures around higher-scoped variables.

> I don't think these functions worth testing

Again, I’d put the burden of proof the other way: assuming it’s quick & easy to write unit tests for these, why not?

FGasper: Closing this thread since it’s probably not worth blocking.

internal/retry/retryer_test.go
huan-Mongo: Make sense! I added a test for pre-v5.

FGasper (approved):
LGTM. Thanks!

PR #13, opened 2024-03-26
REP-4296 Add a `--configFile` parameter.

FGasper:
This allows users to specify the verifier’s configuration via a configuration file rather than CLI parameters.


tfogo:
What's the behavior if you specify an option in the file and on the command line?

FGasper:
@tfogo What I saw is that it amalgamates them both. I don‘t know if we want to specify that behavior, though.

mmcclimon (approved):
LGTM!

tfogo:
Hmmm... it's probably fine, whichever way it works tbh.

tfogo (approved):
LGTM

tdq45gj (approved):
LGTM!

PR #14, opened 2024-03-27
REP-4299 Make `configFile` optional.

FGasper:
PR #13 added a `configFile` parameter. That changeset errantly made that parameter compulsory. This fixes that so that `--configFile` is optional, as we want.

### DIFF
--- a/main/migration_verifier.go
+++ b/main/migration_verifier.go
@@ -153,10 +153,19 @@ func main() {
 	}
 
 	app := &cli.App{
-		Name:   "migration-verifier",
-		Usage:  "verify migration correctness",
-		Flags:  flags,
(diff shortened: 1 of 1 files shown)


jddubois (approved):

mmcclimon:
Hrm; I'm confused as to why this is needed, since what we had in #13 was the same as in mongosync. The difference there is that config file is defined as a non-altsrc flag:

```go
		&cli.StringFlag{
			Name:  "config",
			Usage: "specify the path of the config file",
		},
```

I wonder if that would fix the problem here too; if so, I think it's probably easier to understand than this change (and if nothing else, makes this more similar to mongosync).

It's worth a shot, but this comment isn't blocking; if this works, go for it!

FGasper:
@mmcclimon Ahh, good insight. Let me try that.

FGasper:
@mmcclimon That doesn’t seem to work.

I’ll merge as-is.

PR #15, opened 2024-05-02
REP-2460 Dereference pointers when showing index discrepancies.

FGasper:
Previously we’d show messages like `Mismatch : src: 0xc0a8a7ca7e` when index properties mismatched. This fixes that.


tdq45gj (approved):
LGTM

PR #23, opened 2024-10-17
REP-5141 Add a flag to periodically collect pprof heap usage

tdq45gj:


main/migration_verifier.go
```diff
@@ -146,6 +147,10 @@ func main() {
 			Name:  ignoreReadConcernFlag,
 			Usage: "Use connection-default read concerns rather than setting majority read concern. This option may degrade consistency, so only enable it if majority read concern (the default) doesn’t work.",
 		}),
+		altsrc.NewStringFlag(cli.StringFlag{
+			Name:  pprofInterval,
+			Usage: "Interval to periodically collect pprof profiles",
```
FGasper, line 152: This should give examples or somehow indicate how time units work here.

internal/verifier/pprof.go
```diff
@@ -0,0 +1,45 @@
...
+
+}
+
+func collectHeapUsage() {
+	heapFileName := fmt.Sprintf("heap-%s.out", time.Now().UTC().Format("20060102T150405Z"))
+	heapFile, err := os.Create(heapFileName)
+	defer heapFile.Close()
+
+	if err != nil {
+		panic(err)
```
FGasper, line 38: Panicking seems a bit “loud” for a filesystem failure, but since this is just diagnostic it’s probably OK.

internal/verifier/pprof.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	err = pprof.Lookup("heap").WriteTo(heapFile, 0)
+	if err != nil {
+		heapFile.WriteString(err.Error())
```
FGasper, line 43: Why write the error to the file rather than panicking?

FGasper (changes requested):
A couple small notes.

Thanks!

internal/verifier/pprof.go
tdq45gj: I was thinking about logging and skipping the errors. But it's not intended to be used for production, panicking should be fine.

FGasper (approved):
lgtm

PR #24, opened 2024-10-22
Clarify reporting of change events in progress logs.

FGasper:
3 things cause migration-verifier to recheck a document:
1. document is mismatched between source & destination
2. document is missing on source or destination
3. document changed on source and needs a recheck

Unfortunately, the verifier’s metadata persists 2 and 3 together, such that the verifier’s progress logs can’t differentiate them. The current progress logs, though, report them all as “missing”, which can confuse users.

This changeset updates the logging verbiage to be clearer that the “missing documents” totals are really “missing or changed documents”.


tdq45gj (approved):
LGTM

PR #25, opened 2024-10-23
REP-5140 Report change events in progress logs.

FGasper:
Change events can cause confusion due to how the verifier tallies them alongside missing documents. This changeset tries to add clarity by logging the change events seen during a given generation.


tdq45gj (approved):
LGTM

PR #27, opened 2024-10-29
Reduce memory usage of the migration verifier

tdq45gj:
Generation 0 heap pprof file has shown high memory usage by the `ImportFromCursor` function itself. This PR attempts to improve the memory usage of the migration verifier by:

- removing an extra BSON decode step before storing a document to a doc map in `ImportFromCursor`
- using a byte buffer when constructing map keys in `getMapKey`

### DIFF
--- a/internal/documentmap/documentmap.go
+++ b/internal/documentmap/documentmap.go
@@ -199,8 +199,8 @@ func (m *Map) getMapKey(doc *bson.Raw) MapKey {
 	var keyBuffer bytes.Buffer
 	for _, keyName := range m.fieldNames {
 		value := doc.Lookup(keyName)
-		keyBuffer.WriteString(string(value.Type))
-		keyBuffer.WriteString(string(value.Value))
+		keyBuffer.WriteByte(byte(value.Type))
+		keyBuffer.Write(value.Value)
 	}
 
 	return MapKey(keyBuffer.String())
--- a/internal/documentmap/documentmap.go
+++ b/internal/documentmap/documentmap.go
@@ -199,6 +199,7 @@ func (m *Map) getMapKey(doc *bson.Raw) MapKey {
 	var keyBuffer bytes.Buffer
 	for _, keyName := range m.fieldNames {
 		value := doc.Lookup(keyName)
+		keyBuffer.Grow(1 + len(value.Value))
 		keyBuffer.WriteByte(byte(value.Type))
 		keyBuffer.Write(value.Value)
 	}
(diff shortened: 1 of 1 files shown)


internal/documentmap/documentmap.go
```diff
@@ -103,23 +105,17 @@ func (m *Map) ImportFromCursor(ctx context.Context, cursor *mongo.Cursor) error
...
 
-		var rawDoc bson.Raw
-		err = cursor.Decode(&rawDoc)
-		if err != nil {
-			return err
-		}
-		bytesReturned += (int64)(len(rawDoc))
-
-		m.addDocument(rawDoc)
+		m.addDocument(cursor.Current)
```
FGasper, line 118: nit: Since the copy in `addDocument` is now crucial to correctness, I suggest renaming it `copyAndAddDocument` or `addCopyOfDocument`.

internal/documentmap/documentmap.go
```diff
@@ -200,11 +196,12 @@ func (m *Map) TotalDocsBytes() types.ByteCount {
 
 // called in tests as well as internally
 func (m *Map) getMapKey(doc *bson.Raw) MapKey {
-	key := MapKey("")
+	var keyBuffer bytes.Buffer
 	for _, keyName := range m.fieldNames {
 		value := doc.Lookup(keyName)
-		key += MapKey(value.Type) + MapKey(value.Value)
+		keyBuffer.WriteByte(byte(value.Type))
+		keyBuffer.Write(value.Value)
```
FGasper, line 203: How does this save memory usage?

FGasper (changes requested):
Good find on the extra decode! I’m not sure, though, what we gain from switching to a byte buffer … ?

internal/documentmap/documentmap.go
tdq45gj: I'm not sure but I assume that the old code constructs new strings for each iteration, which is less efficient than continuously appending substrings to a buffer.

FGasper: > less efficient than continuously appending substrings to a buffer

I’m not sure of this. Consider libc’s `realloc` function, which solves the same problem of “extending” a given buffer. Note that it returns a new pointer, which is because there may not be enough room to extend the buffer, in which case a new buffer is allocated and the existing buffer’s contents are copied over.

Extending `keyBuffer` twice here may actually be less efficient than the status quo of extending `key` only once.

FGasper (changes requested):
See note re allocation.

internal/documentmap/documentmap.go
tdq45gj: My understanding is that the status quo does two string concatenations. So it's not just extending `key` once, but it's allocating two new strings each time.

I added `keyBuffer.Grow` to pre-allocate space on the buffer.

FGasper: Status quo allocates two small strings and concatenates them. That’s a smaller haul, I think, than two resizes of a large buffer.

Grow() is probably the way to go here.

FGasper (approved):
LGTM. Hope this helps.

PR #29, opened 2024-10-30
Add GitHub CI

FGasper:


tdq45gj (approved):
I'm a bit surprised that it's not running on evergreen. But I guess we don't have to.

FGasper:
There’s no reason we can’t run both, but this works without additional Evergreen configuration.

PR #34, opened 2024-11-05
REP-5230 Check documents while they’re being fetched.

FGasper:
Previously this tool would, for each partition, fetch all documents for src & dst then compare them. The problem here is that, if any partition happens to be oversize, the verifier might consume so much memory that it crashes.

This changeset solves that problem by making the verifier compare documents as they’re fetched (i.e., in a thread concurrent to the document-fetching threads) and sorting the query results.

For example, assume the source has documents [A B C0 D], and the destination has [A C1 D]. (C0 and C1 mismatch.) One thread will fetch from the source, and another will fetch from the destination. A third will receive documents via channels from the other two threads. Before this changeset, all 7 documents would be cached in memory before they’re compared. With this changeset the checker thread—which holds the documents caches—may work thus:
1. Receive A from source & cache it.
2. Receive A from destination. We have the source’s A, so we can compare them and _discard the cached A_.
3. Receive B from source & cache it.
4. Receive C1 from destination & cache it.
5. (The source-reader thread lags a bit.) Receive D from destination & cache it.
6. Receive C0 from the source. We have the destination’s C, so we can compare them and discard the cached C. They mismatch, so we save that result.
7. Receive D from the source. Compare with the destination’s, and discard the cached one.

At the end, the only cached document will be the source’s B. Since this document remains in the source’s document cache, we know it’s missing on the destination. Note that we only cached at most 3 documents, and that we reduced memory consumption as the fetch proceeded.

This may slightly affect performance because the document-checker thread now constrains the reader threads. At the same time, the CPU-intensive work of comparing the documents now happens alongside the I/O-intensive work of fetching them, which may offset that or even effect a net improvement. (In testing no effect was visible.)


tdq45gj (approved):
LGTM

PR #35, opened 2024-11-06
REP-5218 Roll back tasks as needed when restarting.

FGasper:
The worker threads look for status=added tasks when finding work to do. When it grabs a task it sets status=processing; thus, no other threads will try to grab that task.

If the verifier crashes, though, any tasks that were status=processing need to be reset; otherwise no worker thread will ever pick them up, and the verifier will hang.

This changeset implements that reset of status=processing tasks so that the verifier will be able to finish after a restart.

Specifically:
- If the primary task is found `processing`, this means that the initial listing of namespaces was interrupted. Thus, we now reset the primary task to `added`, and all other tasks are deleted.
- If any collection-verification tasks are found `processing`, that means we never finished writing out the collection’s partitions and checking its metadata. Any existing partition tasks are unusuable, so they get deleted, and the collection-verification task is reset to `added`.
- Any remaining `processing` partition tasks are reset to `added`.

The use of transactions to do these rollbacks means that the verifier’s metadata can no longer be a standalone mongod.


pmeredit:
Commenting here for posterity, this LGTM

internal/verifier/util.go
```diff
@@ -89,3 +91,26 @@ type QueryFilter struct {
 	Namespace string `bson:"namespace"    json:"namespace"`
 	To        string `bson:"to,omitempty" json:"to,omitempty"`
 }
+
+func (verifier *Verifier) doInMetaTransaction(
+	ctx context.Context,
+	todo func(context.Context, mongo.SessionContext) error,
```
autarch, line 97: nit: I think `todo` is not the best name, given how we use `TODO` comments in our code base.

autarch (approved):
LGTM % I don't really have a ton of context on this code base.

PR #36, opened 2024-11-06
REP-5201 Fix change stream iteration.

FGasper:
PR #30 added persistence of the change stream’s resume token but mishandled some of the refactoring by making every change stream iteration loop read all events.

This restores the intended logic: check for context cancellation or the end of the change stream after each event.


tdq45gj (approved):
LGTM

PR #39, opened 2024-11-07
REP-5219 Make migration-verifier process change events in batch.

tdq45gj:


internal/verifier/change_stream.go
```diff
@@ -49,7 +52,7 @@ func (uee UnknownEventError) Error() string {
 
 // HandleChangeStreamEvent performs the necessary work for change stream events that occur during
 // operation.
-func (verifier *Verifier) HandleChangeStreamEvent(ctx context.Context, changeEvent *ParsedEvent) error {
+func (verifier *Verifier) HandleChangeStreamEvent(changeEvent *ParsedEvent) error {
```
FGasper, line 55: Just curious, what was the advantage of doing the buffering here versus closer to the read from the cursor? I was thinking we’d read all the events into a slice then pass that slice to this function (renamed `HandleChangeStreamEvents`).

internal/verifier/change_stream.go
```diff
@@ -98,20 +103,35 @@ func (verifier *Verifier) iterateChangeStream(ctx context.Context, cs *mongo.Cha
...
 				return false, errors.Wrap(err, "failed to decode change event")
 			}
-			err := verifier.HandleChangeStreamEvent(ctx, &changeEvent)
+			err := verifier.HandleChangeStreamEvent(&changeEvent)
 			if err != nil {
 				return false, errors.Wrap(err, "failed to handle change event")
 			}
+
+			eventsRead++
+		}
```
FGasper, line 126: Should there be error handling for cs.Err() here?

internal/verifier/change_stream.go
```diff
@@ -98,20 +103,35 @@ func (verifier *Verifier) iterateChangeStream(ctx context.Context, cs *mongo.Cha
 		return err
 	}
 
-	readOneChangeEvent := func() (bool, error) {
-		gotEvent := cs.TryNext(ctx)
-		if gotEvent {
+	readAndHandleOneChangeEventBatch := func() (bool, error) {
+		eventsRead := 0
+		for hasEventInBatch := true; hasEventInBatch; hasEventInBatch = cs.RemainingBatchLength() > 0 {
```
FGasper, line 108: Checking RemainingBatchLength() before calling TryNext() seems like it risks missing events. I think there has to be a TryNext() before calling RemainingBatchLength() ?

internal/verifier/change_stream.go
```diff
@@ -189,10 +209,14 @@ func (verifier *Verifier) iterateChangeStream(ctx context.Context, cs *mongo.Cha
 }
 
 // StartChangeStream starts the change stream.
-func (verifier *Verifier) StartChangeStream(ctx context.Context) error {
+func (verifier *Verifier) StartChangeStream(ctx context.Context, batchSize *int32) error {
```
FGasper, line 212: Why does this need to be a setting? Aren’t we just taking whatever batch size we get from the source?

internal/verifier/change_stream.go
```diff
@@ -352,3 +376,30 @@ func getClusterTimeFromSession(sess mongo.Session) (primitive.Timestamp, error)
...
+		//
+		// Note that this prevents us from being able to report a meaningful
+		// total data size for noninitial generations in the log.
+		dataSizes := make([]int, len(ids))
+		for i := 0; i < len(ids); i++ {
+			dataSizes[i] = maxBSONObjSize
+		}
+
+		dbName, collName := SplitNamespace(namespace)
+		if err := verifier.insertRecheckDocs(ctx, dbName, collName, ids, dataSizes); err != nil {
```
FGasper, line 397: This approach will cause us to block between each namespace’s events, I think.

Could we instead persist all of the events for all namespaces at once?

internal/verifier/recheck.go
```diff
@@ -91,6 +77,11 @@ func (verifier *Verifier) insertRecheckDocs(
 		verifier.logger.Debug().Msgf("Persisted %d recheck doc(s) for generation %d", len(models), generation)
 	}
 
+	// Silence any duplicate key errors as recheck docs should have existed.
+	if mongo.IsDuplicateKeyError(err) {
```
FGasper, line 81: Dupe key errors can exist alongside other errors.

We should probably instead look for non-simple dupe-key errors so that anything that _isn’t_ a dupe-key will still be handled.

internal/verifier/migration_verifier.go
```diff
@@ -136,6 +136,8 @@ type Verifier struct {
 	globalFilter map[string]any
 
 	pprofInterval time.Duration
+
+	changeEventRecheckBuf ChangeEventRecheckBuffer
```
FGasper, line 140: Suggestion: We shouldn’t need this if we switch to reading all events from the cursor into a slice.

FGasper (changes requested):
Generally this looks good, but I wonder if we can’t simplify it some.

Also the dupe-key handling seems like a point of potential concern. How necessary is that?

internal/verifier/change_stream_test.go
```diff
@@ -193,8 +193,37 @@ func (suite *MultiSourceVersionTestSuite) TestNoStartAtTime() {
...
+		suite.T(),
+		func() bool {
+			return len(suite.fetchVerifierRechecks(ctx, verifier)) == 3
+		},
+		time.Minute,
+		500*time.Millisecond,
+		"the verifier should flush a recheck doc after a batch",
+	)
+	suite.Require().Empty(verifier.changeEventRecheckBuf["testDB.testColl1"])
+	suite.Require().Empty(verifier.changeEventRecheckBuf["testDB.testColl2"])
```
FGasper, line 228: Can we avoid testing verifier internals here? It’d be ideal not to depend on them.

internal/verifier/change_stream.go
tdq45gj: I think the for loop actually calls RemainingBatchLength after calling TryNext(). It's like 
```
for i := 0; i < 10; i++ {
	fmt.Print(i)
}
>> 0123456789
```

internal/verifier/change_stream.go
tdq45gj: I did this for a test to pass in a small batch size.

FGasper: Was that necessary, though?

`getMore` always returns after a relatively short period of inactivity with just whatever events it happens to have seen. So the batch size should be naturally limited to whichever events you’ve triggered … ?

internal/verifier/change_stream.go
tdq45gj: Good point, I made the changes.

internal/verifier/recheck.go
```diff
@@ -36,46 +35,32 @@ func (verifier *Verifier) InsertFailedCompareRecheckDocs(
...
-		ctx, changeEvent.Ns.DB, changeEvent.Ns.Coll, documentIDs, dataSizes)
+	return verifier.insertRecheckDocsWhileLocked(context.Background(),
+		dbNames, collNames, documentIDs, dataSizes)
 }
 
-func (verifier *Verifier) insertRecheckDocsUnderLock(
+func (verifier *Verifier) insertRecheckDocsWhileLocked(
 	ctx context.Context,
-	dbName, collName string, documentIDs []interface{}, dataSizes []int) error {
+	dbNames []string, collNames []string, documentIDs []interface{}, dataSizes []int) error {
```
autarch, line 54: nit: This is a very weird way to format this. I think it'd more readable as one arg per line.

internal/verifier/change_stream.go
```diff
@@ -47,30 +47,53 @@ func (uee UnknownEventError) Error() string {
...
+			dataSizes[i] = maxBSONObjSize
+		default:
+			return UnknownEventError{Event: &changeEvent}
+		}
 	}
+
+	verifier.mux.Lock()
+	defer verifier.mux.Unlock()
+
+	return verifier.insertRecheckDocsWhileLocked(ctx, dbNames, collNames, docIDs, dataSizes)
```
autarch, line 96: I think it'd make more sense to put the locking in the `insertRecheckDocsWhileLocked` method. Right now every caller has to remember to use the mutex properly, which seems like a recipe for mistakes.

internal/verifier/change_stream.go
```diff
@@ -102,20 +125,38 @@ func (verifier *Verifier) iterateChangeStream(ctx context.Context, cs *mongo.Cha
...
+
+			if err := cs.Decode(&changeEventBatch[eventsRead]); err != nil {
+				return false, errors.Wrap(err, "failed to decode change event")
 			}
+
+			eventsRead++
+		}
+
+		if eventsRead > 0 {
+			verifier.logger.Debug().Msgf("Received a batch of %d events", eventsRead)
```
autarch, line 151: Do you also want to log the 0 events case? I don't have enough context to know if that would be useful.

autarch (approved):
LGTM % some small stuff. No need for another review.

internal/verifier/change_stream.go
tdq45gj: I'd prefer not to log if there's no event. We've recommended users to run the m-v with debug log level, so I don't want too many log lines even at debug level.

internal/verifier/change_stream.go
FGasper: I made it this way just this week to accommodate stats. If we’re going to change this I’d rather it’d be in a separate PR since the changes don’t really seem germane.

internal/verifier/change_stream.go
FGasper: I see where you’re breaking out of the loop, but I don’t see anything that reports the error … ?

internal/verifier/change_stream.go
FGasper: I don’t think 0 events is even possible since we wouldn’t be here unless TryNext() returned truthy.

internal/verifier/change_stream.go
```diff
@@ -102,20 +122,38 @@ func (verifier *Verifier) iterateChangeStream(ctx context.Context, cs *mongo.Cha
...
```
FGasper, line 148: nit:
```suggestion
			verifier.logger.Debug().Int("eventsCount", eventsRead).Msgf("Received a batch of events.")
```

internal/verifier/change_stream.go
```diff
@@ -102,20 +122,38 @@ func (verifier *Verifier) iterateChangeStream(ctx context.Context, cs *mongo.Cha
...
+		}
+
+		err := verifier.HandleChangeStreamEvents(ctx, changeEventBatch)
```
FGasper, line 151: Should this handling logic also be under the >0 condition?

internal/verifier/change_stream.go
tdq45gj: I've removed the change event stats from this function in this PR. I think it makes sense to make this change here as well.

internal/verifier/migration_verifier_test.go
```diff
@@ -1459,6 +1466,9 @@ func (suite *MultiDataVersionTestSuite) TestGenerationalRechecking() {
 	suite.Require().NoError(err)
 	// there should be a failure from the src insert
 	suite.Require().Equal(VerificationStatus{TotalTasks: 1, FailedTasks: 1}, *status)
+
+	checkContinueChan <- struct{}{}
+	require.NoError(suite.T(), errGroup.Wait())
```
FGasper, line 1471: Nice find!

FGasper (approved):
LGTM. Thanks!

PR #41, opened 2024-11-14
REP-5277 Parallelize persistence of rechecks.

FGasper:
We receive events and enqueue rechecks in the same thread. When there are 1,000s of rechecks to enqueue at once, though, we need faster throughput than a single thread will allow.

This changeset fixes that by parallelizing the enqueueing of rechecks.

### DIFF
--- a/internal/verifier/recheck.go
+++ b/internal/verifier/recheck.go
@@ -3,15 +3,20 @@ package verifier
 import (
 	"context"
 
+	"github.com/10gen/migration-verifier/internal/retry"
 	"github.com/10gen/migration-verifier/internal/types"
+	"github.com/pkg/errors"
+	"github.com/samber/lo"
 	"go.mongodb.org/mongo-driver/bson"
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
+	"golang.org/x/sync/errgroup"
 )
 
 const (
-	recheckQueue   = "recheckQueue"
-	maxBSONObjSize = 16 * 1024 * 1024
+	recheckQueue                  = "recheckQueue"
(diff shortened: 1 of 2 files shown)


tdq45gj (approved):
LGTM % one suggestion

internal/verifier/recheck.go
```diff
@@ -63,38 +84,78 @@ func (verifier *Verifier) insertRecheckDocs(
...
-		// same _id as the one that the 1st insert already created.
-		filterDoc := bson.D{{"_id", pk}}
+	eg, groupCtx := errgroup.WithContext(ctx)
 
-		recheckDoc := RecheckDoc{
-			PrimaryKey: pk,
-			DataSize:   dataSizes[i],
-		}
+	for _, curThreadIndexes := range indexesPerThread {
+		curThreadIndexes := curThreadIndexes
```
jtcovan, line 93: Why do we redefine the var?

jtcovan (approved):
LGTM % question around redefining var

internal/verifier/recheck.go
FGasper: The linter flagged it as one of those cases where you’re supposed to redefine the var to prevent loop-variable-in-goroutine problems.

internal/verifier/recheck.go
```diff
@@ -51,6 +60,18 @@ func (verifier *Verifier) InsertFailedCompareRecheckDocs(
 	)
 }
 
+// This will split the given slice into *roughly* the given number of chunks.
+// It may end up being more or fewer, but it should be pretty close.
+func splitToChunks[T any, Slice ~[]T](elements Slice, numChunks int) []Slice {
+	elsPerChunk := len(elements) / numChunks
```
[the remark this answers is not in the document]
FGasper: done

PR #43, opened 2024-11-15
REP-5283 Report change event document sizes accurately.

FGasper:
Previously all change events’ document sizes were recorded “pessimistically” as 16 MiB. This helped to avoid OOMs. It came at a cost, though: when the recheck queue is converted to recheck tasks, those tasks are sized so as to approximate the configured partition size. Thus, if the partition size was 400 MiB (the default), only 25 change events could fit into a recheck task. If there are 250,000 pending rechecks—not unfeasible for a large, busy data set after generation 0—that’s 10,000 tasks to create and perform, which is inefficient.

PR #34 all but eliminates the OOMs, which undercuts that “pessimism”’s benefit. It makes more sense now to allow for the possibility of large recheck tasks in order to minimize the number of tasks. Moreover, we can get pretty good confidence about document sizes from change events anyway:
- Insert & replace events always include the `fullDocument`.
- Update events can be configured to include the current `fullDocument`.
- Delete events refer to a document that probably no longer exists, so we can safely estimate its size to be “small”.

This changeset, then:
1. configures the change stream to include `fullDocument` in update events, and
2. records document sizes from the change event.

### DIFF
--- a/internal/verifier/change_stream.go
+++ b/internal/verifier/change_stream.go
@@ -108,7 +108,7 @@ func (verifier *Verifier) GetChangeStreamFilter() []bson.D {
 	if len(verifier.srcNamespaces) == 0 {
 		pipeline = []bson.D{{{"$match", bson.D{{"ns.db", bson.D{{"$ne", verifier.metaDBName}}}}}}}
 	} else {
-		filter := []bson.D{}
+		filter := bson.A{}
 		for _, ns := range verifier.srcNamespaces {
 			db, coll := SplitNamespace(ns)
 			filter = append(filter, bson.D{{"ns", bson.D{{"db", db}, {"coll", coll}}}})
--- a/internal/verifier/change_stream.go
+++ b/internal/verifier/change_stream.go
@@ -102,6 +102,13 @@ func (verifier *Verifier) HandleChangeStreamEvents(ctx context.Context, batch []
 	return verifier.insertRecheckDocs(ctx, dbNames, collNames, docIDs, dataSizes)
 }
 
+// GetChangeStreamFilter returns an aggregation pipeline that filters
+// namespaces as per configuration.
+//
+// NB: Ideally we could make the change stream give $bsonSize(fullDocument)
+// and omit fullDocument, but $bsonSize was new in MongoDB 4.4, and we still
+// want to verify migrations from 4.2. fullDocument is unlikely to be a
+// bottleneck anyway.
 func (verifier *Verifier) GetChangeStreamFilter() []bson.D {
 	if len(verifier.srcNamespaces) == 0 {
 		return []bson.D{{bson.E{"$match", bson.D{{"ns.db", bson.D{{"$ne", verifier.metaDBName}}}}}}}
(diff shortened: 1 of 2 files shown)


internal/verifier/change_stream.go
```diff
@@ -15,12 +15,15 @@ import (
...
 
+const fauxDocSizeForDeleteEvents = 1024
+
 // ParsedEvent contains the fields of an event that we have parsed from 'bson.Raw'.
 type ParsedEvent struct {
 	ID          interface{}          `bson:"_id"`
 	OpType      string               `bson:"operationType"`
 	Ns          *Namespace           `bson:"ns,omitempty"`
 	DocKey      DocKey               `bson:"documentKey,omitempty"`
+	DocSize     *int                 `bson:"fullDocument,omitempty"`
```
tdq45gj, line 26: The BSON tag should probably be `documentSize`.

internal/verifier/change_stream.go
```diff
@@ -98,16 +103,52 @@ func (verifier *Verifier) HandleChangeStreamEvents(ctx context.Context, batch []
...
+				{"$addFields", bson.D{
+					{"documentSize", bson.D{
+						{"$cond", bson.D{
+							{"if", bson.D{
+								{"$ne", bson.A{
+									"missing",
+									bson.D{{"$type", "$fullDocument"}},
+								}},
+							}}, // fullDocument exists
+							{"then", bson.D{{"$bsonSize", "$fullDocument"}}},
```
tdq45gj, line 135: `$bsonSize` is added in v.4.4. Can we fall back to the old slower logic for v4.2 and below?

internal/verifier/change_stream.go
```diff
@@ -98,16 +103,52 @@ func (verifier *Verifier) HandleChangeStreamEvents(ctx context.Context, batch []
...
+		[]bson.D{
+
+			// Add a documentSize field.
+			{
+				{"$addFields", bson.D{
+					{"documentSize", bson.D{
+						{"$cond", bson.D{
+							{"if", bson.D{
+								{"$ne", bson.A{
+									"missing",
```
tdq45gj, line 131: Do we need to check the fullDocument missing case? `$bsonSize` returns 0 on null value.

tdq45gj (changes requested):
This looks good overall. I wonder if it's necessary to check the fullDocument missing case since.

internal/verifier/change_stream.go
FGasper: oops! Good catch.

internal/verifier/change_stream.go
FGasper: Thank you for checking this. I checked as far back as 5.0 but not earlier.

That makes me think we should just do `fullDocument`. I don’t think this is where any bottlenecks are anyway.

Thoughts?

internal/verifier/change_stream.go
FGasper: IMO we should avoid “magic values” when possible. I realize that 0 often serves in that role, but if we can avoid that I think it’s a small code-quality win.

mtrussotto (approved):
updateLookup is fairly expensive (not the network cost, but the query), but I can believe the savings from better partitioning outweigh that.

PR #44, opened 2024-11-15
Improve logging statements’ accuracy and reporting of data points.

FGasper:
This changeset improves several logs by:
- reporting their data points as structured data rather than phrase components
- including the task ID (“PrimaryKey” in code) to facilitate correlation among logs and metadata
- adding a few other data points and adding new log events such as task-finished


tdq45gj (approved):
LGTM

PR #45, opened 2024-11-15
Ensure parallelization of rechecks by capping recheck task size.

FGasper:
Presently the only limit on a recheck task’s size is BSON’s document-size limit (i.e., 16 MiB). Thus, a single task can feasibly include every recheck. If there are millions of rechecks, that means only 1 thread is doing all of them, which increases the likelihood that migration-verifier would fail to keep pace with the source workload.

This changeset addresses this problem by capping the recheck task size according to the number of pending rechecks and the number of workers.

### DIFF
--- a/internal/reportutils/reportutils.go
+++ b/internal/reportutils/reportutils.go
@@ -15,10 +15,11 @@ import (
 
 const decimalPrecision = 2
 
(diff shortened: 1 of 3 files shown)


tdq45gj (approved):
LGTM

PR #46, opened 2024-11-18
Add a timeout when reading cursor, and fail immediately.

FGasper:
In certain cases reads from the source or destination seem to hang. This changeset makes two changes to address that:
1. A timer is added to detect hung reads. It is hard-coded at 10 minutes, which seems conservative enough.
2. The comparer thread fails immediately if it finds a failure from the source. Previously it would read from the destination if the source failed. (This _could_ be the case of the apparent hangs, but how that could be seems unclear.)

### DIFF
--- a/internal/verifier/compare.go
+++ b/internal/verifier/compare.go
@@ -3,6 +3,7 @@ package verifier
 import (
 	"bytes"
 	"context"
+	"time"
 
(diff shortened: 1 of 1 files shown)


tdq45gj (approved):
LGTM

PR #47, opened 2024-11-19
REP-5299 Finalize change stream with a timestamp.

FGasper:
Previously to finalize its change stream the verifier would just TryNext() until no responses arrived. This works fine for unsharded clusters, but for sharded clusters it can potentially miss events.

This changeset adopts more “tried-and-true” logic: grab a timestamp via `appendOplogNote`, then iterate the change stream until the stream’s resume token’s timestamp meets or exceeds that timestamp.

The timestamp-grabbing logic is unconditionally retryable, which requires an update to the retryer module to implement. This changeset refactors that module a bit to prevent further duplication.


mbson/bson_raw.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"github.com/pkg/errors"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/x/bsonx/bsoncore"
+)
+
+// RawLookup combines bson.Raw’s LookupErr method with an additional
+// unmarshal step. The result is a convenient way to extract values from
+// bson.Raw. The returned boolean indicates whether the value was found.
+func RawLookup[T any](doc bson.Raw, dest *T, keys ...string) (bool, error) {
```
mmcclimon, line 12: (For the record, I still find this function name extremely confusing.)

internal/verifier/logging_setup.go
```diff
@@ -29,3 +29,15 @@ func getLogWriter(logPath string) io.Writer {
 
 	return zerolog.SyncWriter(writer)
 }
+
+func getLoggerAndWriter(logPath string) (*logger.Logger, io.Writer) {
+	writer := getLogWriter(logPath)
+
+	consoleWriter := zerolog.ConsoleWriter{
```
mmcclimon, line 36: I'm not sure what this is doing here.  It feels like we probably shouldn't do this unconditionally. If we're logging to a file, using the console writer would spew a bunch of color codes into the log, wouldn't it?

internal/verifier/migration_verifier.go
```diff
@@ -315,15 +313,7 @@ func (verifier *Verifier) SetPartitionSizeMB(partitionSizeMB uint32) {
 }
 
 func (verifier *Verifier) SetLogger(logPath string) {
-	writer := getLogWriter(logPath)
-	verifier.writer = writer
-
-	consoleWriter := zerolog.ConsoleWriter{
-		Out:        writer,
-		TimeFormat: timeFormat,
-	}
```
mmcclimon, line 324: Oh I see, we were already going this elsewhere. Weird, but ok.

internal/verifier/change_stream.go
```diff
@@ -140,37 +144,46 @@ func (verifier *Verifier) iterateChangeStream(ctx context.Context, cs *mongo.Cha
...
 			if err := cs.Decode(&changeEventBatch[eventsRead]); err != nil {
-				return false, errors.Wrap(err, "failed to decode change event")
+				return errors.Wrap(err, "failed to decode change event")
 			}
 
 			eventsRead++
 		}
 
-		if eventsRead > 0 {
-			verifier.logger.Debug().Int("eventsCount", eventsRead).Msgf("Received a batch of events.")
```
mmcclimon, line 166: (Just an observation that this log line has disappeared in the shuffle; I have no opinion on whether that's good or bad.)

internal/verifier/clustertime.go
```diff
@@ -0,0 +1,141 @@
...
+const opTimeKeyInServerResponse = "operationTime"
+
+// GetClusterTime returns the remote cluster’s cluster time.
+// In so doing it creates a new oplog entry across all shards.
+// The “note” will go into the cluster’s oplog, so keep it
+// short but meaningful.
+func GetClusterTime(
+	ctx context.Context,
+	logger *logger.Logger,
+	client *mongo.Client, // could change to type=any if we need
```
mmcclimon, line 26: I don't understand this comment; surely we must always have a mongo client, because we're calling mongo-shaped things on it!

internal/verifier/clustertime.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		// This isn't serious enough even to warn on, so leave it at info-level.
+		logger.Info().Err(err).
+			Msg("Failed to append oplog note; change stream may need extra time to finish.")
+	}
+
+	return optime, nil
+}
+
+func runAppendOplogNote(
```
mmcclimon, line 79: I feel like this would be easier to read if it was two different functions: one that's the no-write version (the current doomed=true), and another that's the actutal-write version. I find all the if statements here difficult to follow, plus "doomed" is a weird parameter name (and boolean parameters basically always smell), and if this were `getCurrentTime` and `advanceCurrentTime` or something.

internal/verifier/integration_test_suite.go
```diff
@@ -94,6 +94,19 @@ func (suite *IntegrationTestSuite) SetupTest() {
 		dbname,
 	)
 
+	for _, client := range []*mongo.Client{suite.srcMongoClient, suite.dstMongoClient} {
+		dbNames, err := client.ListDatabaseNames(ctx, bson.D{})
+		suite.Require().NoError(err, "should list database names")
+		for _, dbName := range dbNames {
+			if strings.Index(dbName, suite.DBNameForTest()) == 0 {
```
mmcclimon, line 101: (nit)
```suggestion
			if strings.HasPrefix(dbName, suite.DBNameForTest()) {
```

mmcclimon (changes requested):
I think the general shape here is fine; I left a handful of things mostly for readability. Thanks!

internal/verifier/logging_setup.go
FGasper: This is moved code.

tdq45gj (changes requested):
This looks good. I left some comments about logging and error messages.

internal/verifier/clustertime.go
FGasper: Now that you’ve made me look at this again :-P, I find I still don’t fully grok why a 2nd `appendOplogNote` is necessary to prevent a delay from the shards in outputting their events.

I’m just going to simplify it to a single `appendOplogNote`. It seems like that’s all that should be needed anyway, and if I’m wrong, a 10-second wait is miniscule in the grand scheme.

internal/retry/retry.go
```diff
@@ -273,25 +276,42 @@ func (r *Retryer) shouldRetryWithSleep(
...
 		}
+
+		event.Msg("Not retrying on error because it appears not to be transient.")
+
+		return false
+	}
+
+	retryLogEvent := logger.Warn().
+		Err(err).
+		Stringer("duration", sleepTime)
```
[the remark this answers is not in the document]
FGasper: That’s what this does: it’ll say `4s` if it’s 4 seconds, etc.

internal/verifier/clustertime.go
```diff
@@ -0,0 +1,141 @@
...
+	err := resp.Err()
+
+	if doomed {
+		// We expect an error here; if we didn't get one then something is
+		// amiss on the server.
+		if err == nil {
+			return primitive.Timestamp{}, errors.Errorf("server request unexpectedly succeeded: %v", cmd)
+		}
+
+		if !util.IsStaleClusterTimeError(err) {
```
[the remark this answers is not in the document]
FGasper: I’ve removed the StaleClusterTime consideration.

internal/verifier/clustertime.go
```diff
@@ -0,0 +1,141 @@
...
+			return primitive.Timestamp{}, errors.Errorf(
+				"unexpected error (expected StaleClusterTime) from request (%v): %v",
+				err,
+				cmd,
```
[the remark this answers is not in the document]
FGasper: This is now removed.

internal/verifier/logging_setup.go
FGasper: This is just moved code.

mbson/bson_raw.go
FGasper: Noted … but the functionality seems just too useful not to bring over.

internal/verifier/check.go
```diff
@@ -66,6 +77,12 @@ func (verifier *Verifier) waitForChangeStream() error {
 func (verifier *Verifier) CheckWorker(ctx context.Context) error {
 	verifier.logger.Debug().Msgf("Starting %d verification workers", verifier.numWorkers)
 	ctx, cancel := context.WithCancel(ctx)
+
+	if verifier.workerSleepDelayMillis == 0 {
+		verifier.logger.Info().
+			Msg("Worker sleep delay is zero. Only tests should do this.")
```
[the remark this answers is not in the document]
FGasper: Hm. Changing that is outside scope here, but I question why a user would need to customize that. I’ll remove this log message.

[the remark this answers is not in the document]
tdq45gj: I'm just saying that it's a possible user setting, not proposing to change the flag 😄

tdq45gj (approved):
LGTM

internal/verifier/clustertime.go
mmcclimon: Even better! This is way easier to follow now, thanks.

internal/verifier/migration_verifier.go
```diff
@@ -228,13 +231,39 @@ func (verifier *Verifier) SetFailureDisplaySize(size int64) {
...
+		}
+
+		verifier.writesOffTimestamp = &finalTs
+
+		verifier.mux.Unlock()
+
+		// This has to happen under the lock because the change stream
+		// might be inserting docs into the recheck queue, which happens
+		// under the lock.
+		verifier.changeStreamWritesOffTsChan <- finalTs
```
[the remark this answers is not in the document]
FGasper: Oof, a mis-worded comment. It should read “This has to happen outside the lock …”

PR #49, opened 2024-11-22
REP-5319 Only create partition-verification tasks in generation 0.

FGasper:
When the verifier finds a metadata mismatch it enqueues a recheck of that metadata in the next generation.

This is fine, but since metadata-check tasks also partition documents & create document-checking tasks for those  partitions, what’s been happening is that any time some collection metadata property (e.g., an index) mismatches, tasks are enqueued to recheck all of the collection’s documents.

Since the partitioning stuff is only appropriate for generation 0, this changeset restricts that behavior accordingly.

For convenience this also adds mslices.Of from mongosync’s mongo-go.


tdq45gj (approved):
LGTM

PR #50, opened 2024-11-22
REP-5274 Compare full index specs rather than just a subset.

FGasper:
Previously migration-verifier only checked the parts of index specs that go in a mongo.IndexSpecification. That struct is incomplete, though; it doesn’t, for example, include partial index filters. Thus, migration-verifier didn’t compare those parts of an index specification.

This changeset alters the index-checking logic to check full index specifications. Note that this comparison is rather “coarse” and may flag some specifications as mismatching when they actually produce the same index. For example, `{foo: 1}` and `{foo: 2}` actually produce the same index, but migration-verifier will flag them as mismatches. Eventually we hope to fix that, but for now false positives are safer than incomplete checks.

This also replaces use of mongo.CollectionSpecification with a struct that allows detection of unrecognized fields. Thus we’ll avoid a similar problem for collection specifications. 

This also copies over mongosync’s Option implementation.


option/bson.go
```diff
@@ -0,0 +1,49 @@
...
+	if !exists {
+		return bson.MarshalValue(primitive.Null{})
+	}
+
+	return bson.MarshalValue(val)
+}
+
+// UnmarshalBSONValue implements bson.ValueUnmarshaler.
+func (o *Option[T]) UnmarshalBSONValue(bType bsontype.Type, raw []byte) error {
+
```
tdq45gj, line 22: nit:
```suggestion
```

internal/verifier/migration_verifier.go
```diff
@@ -875,64 +866,74 @@ func (verifier *Verifier) compareCollectionSpecifications(srcNs string, dstNs st
...
-
-	if !reflect.DeepEqual(srcSpec.ExpireAfterSeconds, dstSpec.ExpireAfterSeconds) {
-		results = append(results, VerificationResult{
-			NameSpace: dstSpec.Namespace,
-			Cluster:   ClusterTarget,
-			ID:        dstSpec.Name,
-			Field:     "ExpireAfterSeconds",
-			Details:   Mismatch + fmt.Sprintf(" : src: %s, dst: %s", nilableToString(srcSpec.ExpireAfterSeconds), nilableToString(dstSpec.ExpireAfterSeconds))})
+	// Next check to see if the only differences are type differences.
+	// If we didn’t support pre-v5 servers we could use a $documents aggregation,
```
tdq45gj, line 876: I think we can reasonably limit the metadata cluster to post-v5 versions, but probably doesn't worth to do only for `$document`.

tdq45gj (approved):
LGTM

internal/verifier/migration_verifier.go
FGasper: ahhh, right! I’d overlooked that this is the metadata cluster. Totally worth it, then, IMO.

PR #51, opened 2024-11-22
Suppress color output when writing to something other than a terminal.

FGasper:
It’s awkward to parse migration-verifier logs in editors and pagers that don’t handle ANSI escape sequences. This changeset alters migration-verifier’s logging so that it only colorizes when outputting to a terminal.

Special provision is made here to preserve colorization of test output.


jddubois (approved):
approving for test, will revoke in a second

PR #52, opened 2024-11-22
REP-5318 Retry on change stream failures.

FGasper:
# REVIEW NOTE

It probably makes sense to review this PR’s 2 commits separately.

-----

This is necessary for testing migration-verifier’s change stream in mongosync’s nondeterministic tests. It entails a small refactor of the change stream code so that both the change stream’s creation and iteration happen under a retryer.

Additionally, this makes GetNewClusterTime() return an actual _new_ cluster time rather than returning the current one. This simplifies reasoning about the change stream’s handling of the writesOff timestamp.

### DIFF
--- a/internal/verifier/change_stream.go
+++ b/internal/verifier/change_stream.go
@@ -226,7 +226,7 @@ func (verifier *Verifier) iterateChangeStream(ctx context.Context, cs *mongo.Cha
 					break
 				}
 
-				if curTs.After(writesOffTs) {
+				if !curTs.Before(writesOffTs) {
 					verifier.logger.Debug().
 						Interface("currentTimestamp", curTs).
 						Interface("writesOffTimestamp", writesOffTs).
--- a/internal/verifier/change_stream.go
+++ b/internal/verifier/change_stream.go
@@ -6,8 +6,11 @@ import (
 	"time"
 
 	"github.com/10gen/migration-verifier/internal/keystring"
+	"github.com/10gen/migration-verifier/internal/retry"
+	"github.com/10gen/migration-verifier/internal/util"
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog"
+	"github.com/samber/mo"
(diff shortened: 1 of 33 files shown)


internal/verifier/change_stream.go
```diff
@@ -322,40 +316,95 @@ func (verifier *Verifier) StartChangeStream(ctx context.Context) error {
...
+
+					defer srcChangeStream.Close(ctx)
+
+					if parentThreadWaiting {
+						resultChan <- mo.Ok(startTs)
+						close(resultChan)
+						parentThreadWaiting = false
+					}
+
+					return verifier.iterateChangeStream(ctx, srcChangeStream)
```
tdq45gj: I think we need to pass `retry.Info` into `iterateChangeStream` and reset the retryer by calling `IterationSuccess`.

internal/verifier/change_stream.go
```diff
@@ -322,40 +316,95 @@ func (verifier *Verifier) StartChangeStream(ctx context.Context) error {
...
+		startTs = clusterTime
 	}
 
+	return srcChangeStream, startTs, nil
+}
+
+// StartChangeStream starts the change stream.
+func (verifier *Verifier) StartChangeStream(ctx context.Context) error {
+	// Result seems a bit simpler than messing with 2 separate channels.
+	resultChan := make(chan mo.Result[primitive.Timestamp])
```
tdq45gj, line 363: I'm a bit confused. Could you clarify what result does this resultChan expect?

tdq45gj: Does it hold the result from `createChangeStream`? Could we use a more descriptive name such as `createChangeStreamResultChan`?

internal/verifier/change_stream.go
FGasper: done

tdq45gj (approved):
LGTM

PR #53, opened 2024-11-22
REP-5317 Add destination change stream

tdq45gj:
This PR adds a destination change stream to the MV. It also adds a `ChangeStreamReader` struct that reads change events from either source or destination. Change events are handled in change event handlers that're goroutines started for each of the change stream reader.


internal/verifier/change_stream.go
```diff
@@ -52,8 +53,81 @@ func (uee UnknownEventError) Error() string {
...
 
+type ChangeStreamReader struct {
+	readerType clusterType
+
+	lastChangeEventTime *primitive.Timestamp
+	logger              *logger.Logger
+	namespaces          []string
+	metaDBName          string
+
+	metaClient    *mongo.Client
```
FGasper, line 64: A change stream “reader” shouldn’t ordinarily need to access the metadata. Can this be separated out?

If we do need to keep this here, this might as well be `metaDB *mongo.Database` and combine metaDBName and metaClient.

internal/verifier/change_stream.go
```diff
@@ -52,8 +53,81 @@ func (uee UnknownEventError) Error() string {
...
+	ChangeStreamErrChan         chan error
+	ChangeStreamDoneChan        chan struct{}
+
+	startAtTs *primitive.Timestamp
+}
+
+func (verifier *Verifier) initializeChangeStreamReaders() {
+	verifier.srcChangeStreamReader = &ChangeStreamReader{
+		readerType:                  srcReaderType,
+		lastChangeEventTime:         nil,
```
FGasper, line 80: nit: no need to note this (or other zero-values)

internal/verifier/change_stream.go
```diff
@@ -232,17 +327,17 @@ func (verifier *Verifier) iterateChangeStream(
 		case <-ctx.Done():
 			return ctx.Err()
 
-		// If the changeStreamEnderChan has a message, the user has indicated that
-		// source writes are ended. This means we should exit rather than continue
+		// If the ChangeStreamEnderChan has a message, the user has indicated that
+		// source and destination writes are ended. This means we should exit rather than continue
```
FGasper, line 331: I don’t know that the “and destination” part is gainful here. It might suggest that user writes to the destination are expected during a migration, which they definitely aren’t.

internal/verifier/change_stream.go
```diff
@@ -232,17 +327,17 @@ func (verifier *Verifier) iterateChangeStream(
...
-			verifier.logger.Debug().
+		case writesOffTs := <-csr.ChangeStreamWritesOffTsChan:
+			csr.logger.Debug().
 				Interface("writesOffTimestamp", writesOffTs).
 				Msg("Change stream thread received writesOff timestamp. Finalizing change stream.")
 
 			gotwritesOffTimestamp = true
 
-			// Read all change events until the source reports no events.
+			// Read all change events until the source / destination reports no events.
```
FGasper, line 340: This log is inaccurate upstream; we might as well fix it here.

Maybe:
> Read change events until the stream reaches the writesOffTs.

internal/verifier/check.go
```diff
@@ -85,9 +85,12 @@ func (verifier *Verifier) CheckWorker(ctx context.Context) error {
 
 	for {
 		select {
-		case err := <-verifier.changeStreamErrChan:
+		case err := <-verifier.srcChangeStreamReader.ChangeStreamErrChan:
 			cancel()
-			return errors.Wrap(err, "change stream failed")
+			return errors.Wrapf(err, "got an error from %s", verifier.srcChangeStreamReader)
```
FGasper, line 90: Do you prefer “got an error” to “failed” here? The latter seems (to me, at least) like stronger, more concise wording.

Maybe:
```
errors.Wrapf(err, "%s failed", verifier.srcChangeStreamReader)
```

internal/verifier/check.go
```diff
@@ -85,9 +85,12 @@ func (verifier *Verifier) CheckWorker(ctx context.Context) error {
...
+		case err := <-verifier.dstChangeStreamReader.ChangeStreamErrChan:
+			cancel()
+			return errors.Wrapf(err, "got an error from %s", verifier.dstChangeStreamReader)
```
FGasper, line 93: same q re “got an error”

internal/verifier/check.go
```diff
@@ -166,19 +170,30 @@ func (verifier *Verifier) CheckDriver(ctx context.Context, filter map[string]any
...
+	if verifier.dstChangeStreamReader.changeStreamRunning {
+		verifier.logger.Debug().Msg("Check: Destination change stream already running.")
+	} else {
+		verifier.logger.Debug().Msg("Destination change stream not running; starting change stream")
+
+		err = verifier.dstChangeStreamReader.StartChangeStream(ctx)
+		if err != nil {
+			return errors.Wrap(err, "failed to start change stream on destination")
+		}
+		verifier.StartChangeEventHandler(ctx, verifier.dstChangeStreamReader)
```
FGasper, line 194: Since ChangeStreamReader has a String() method, you could just loop over the src & dst change streams rather than duplicating this logic.

FGasper (changes requested):
In broad strokes this looks good.

I’ve noted some small things, mostly nits.

Can there also be a test added against the scenario that we envisioned?
- Update A on source
- Update ~A on source
- verification reports success
- “misapply” updates A/~A on destination

… and confirm that we get a failure.

internal/verifier/change_stream_test.go
```diff
@@ -454,6 +460,68 @@ func (suite *IntegrationTestSuite) TestCreateForbidden() {
...
+	coll1 := db.Collection("dstColl1")
+	coll2 := db.Collection("dstColl2")
+
+	for _, coll := range mslices.Of(coll1, coll2) {
+		suite.Require().NoError(db.CreateCollection(ctx, coll.Name()))
+	}
+
+	verifier := suite.BuildVerifier()
+	verifier.SetSrcNamespaces([]string{"srcDB.srcColl1", "srcDB.srcColl2"})
+	verifier.SetDstNamespaces([]string{"dstDB.dstColl1", "dstDB.dstColl2"})
```
FGasper, line 476: Can these use suite.DBNameForTest() rather than hard-coding DB names?

internal/verifier/change_stream.go
```diff
@@ -52,8 +53,81 @@ func (uee UnknownEventError) Error() string {
...
+	readerType clusterType
+
+	lastChangeEventTime *primitive.Timestamp
+	logger              *logger.Logger
+	namespaces          []string
+	metaDBName          string
+
+	metaClient    *mongo.Client
+	watcherClient *mongo.Client
+	buildInfo     *util.BuildInfo
```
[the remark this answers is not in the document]
tdq45gj: I don't think they will. They're pointers just to be consistent with the `verifier` struct.

internal/verifier/change_stream.go
tdq45gj: I think it means that the migration tool has stopped/committed. I'll clarify in this comment.

internal/verifier/check.go
tdq45gj: Agreed.

internal/verifier/change_stream.go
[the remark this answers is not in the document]
FGasper: The pointer in `verifier` is a bit of a mystery, but I think it’s there because when that struct is initialized the BuildInfo isn’t loaded yet. So it’s “optional” in that way.

That’s not the case here, though: BuildInfo is _always_ available here. While consistency is a good thing on its own, maybe there’s more benefit in avoiding the notion that BuildInfo is “optional” here?

internal/verifier/migration_verifier.go
```diff
@@ -76,6 +76,13 @@ const (
 	notOkSymbol = "\u2757" // heavy exclamation mark symbol
 )
 
+type clusterType string
+
+const (
+	srcReaderType clusterType = "source"
```
FGasper, line 82: “cluster type” seems relatively non-descript: “type” often means topology, for example.

Maybe `whichCluster`?

internal/verifier/migration_verifier.go
```diff
@@ -240,35 +242,48 @@ func (verifier *Verifier) WritesOff(ctx context.Context) error {
...
-	if verifier.writesOffTimestamp == nil {
-		verifier.logger.Debug().Msg("Change stream still running. Signalling that writes are done.")
+	verifier.logger.Debug().Msg("Signalling that writes are done.")
 
-		finalTs, err := GetNewClusterTime(
-			ctx,
-			verifier.logger,
-			verifier.srcClient,
-		)
+	srcFinalTs, err := GetNewClusterTime(
```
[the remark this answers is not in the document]
FGasper: > I don't think GetNewClusterTime needs to be under the lock.

What benefit do we realize from changing this, though?

internal/verifier/migration_verifier.go
```diff
@@ -239,35 +241,48 @@ func (verifier *Verifier) WritesOff(ctx context.Context) error {
...
-		// might be inserting docs into the recheck queue, which happens
-		// under the lock.
-		select {
-		case verifier.changeStreamWritesOffTsChan <- finalTs:
-		case err := <-verifier.changeStreamErrChan:
-			return errors.Wrap(err, "tried to send writes-off timestamp to change stream, but change stream already failed")
-		}
-	} else {
-		verifier.mux.Unlock()
+	// This has to happen outside the lock because the change stream
```
FGasper, line 273: There are 2 change streams … can you amend this comment accordingly?

internal/verifier/migration_verifier_test.go
```diff
@@ -1545,6 +1553,82 @@ func (suite *IntegrationTestSuite) TestVerifierWithFilter() {
...
+	suite.Require().NoError(srcDB.CreateCollection(ctx, collName))
+	suite.Require().NoError(dstDB.CreateCollection(ctx, collName))
+
+	verifier := suite.BuildVerifier()
+	runner := RunVerifierCheck(ctx, suite.T(), verifier)
+
+	// Dry run generation 0 to make sure change stream reader is started.
+	suite.Require().NoError(runner.AwaitGenerationEnd())
+
+	suite.Require().NoError(runner.StartNextGeneration())
```
FGasper, line 1585: Should the generation start after the inserts? It seems like that’d be a bit less race-prone.

FGasper (changes requested):
I’ve left a few more small comments. I think this is quite close! Please ping me when you’ve had a chance to look.

Thanks!

internal/verifier/change_stream.go
```diff
@@ -52,8 +53,79 @@ func (uee UnknownEventError) Error() string {
...
+
+	metaDB        *mongo.Database
+	watcherClient *mongo.Client
+	clusterInfo   util.ClusterInfo
+
+	changeStreamRunning  bool
+	ChangeEventBatchChan chan []ParsedEvent
+	WritesOffTsChan      chan primitive.Timestamp
+	ErrChan              chan error
+	DoneChan             chan struct{}
```
FGasper, line 71: nit: Do these need to be exported?

internal/verifier/change_stream.go
```diff
@@ -185,6 +280,15 @@ func (verifier *Verifier) readAndHandleOneChangeEventBatch(
 			return errors.Wrapf(err, "failed to decode change event to %T", changeEventBatch[eventsRead])
 		}
 
+		csr.logger.Trace().Msgf("%s received a change event: %v", csr, changeEventBatch[eventsRead])
+		fmt.Printf("%d %d\n", changeEventBatch[eventsRead].ClusterTime.T, changeEventBatch[eventsRead].ClusterTime.I)
+
+		if changeEventBatch[eventsRead].ClusterTime != nil &&
```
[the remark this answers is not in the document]
FGasper: I feel like if a missing `clusterTime` ever happened that’d be something we should at least warn about, if not fail outright. But that’d be more apropos to a separate PR.

internal/verifier/change_stream.go
```diff
@@ -194,15 +298,11 @@ func (verifier *Verifier) readAndHandleOneChangeEventBatch(
 		return nil
 	}
 
-	err := verifier.HandleChangeStreamEvents(ctx, changeEventBatch)
-	if err != nil {
-		return errors.Wrap(err, "failed to handle change events")
-	}
-
+	csr.ChangeEventBatchChan <- changeEventBatch
```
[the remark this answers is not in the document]
FGasper: OK, yeah it works around the circular dependency.

Ideally I’d us to solve that by breaking apart Verifier, but that may be a bigger change than the present “mini-project” warrants.

internal/verifier/migration_verifier.go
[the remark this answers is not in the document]
FGasper: My concern is that this is fairly brittle stuff. We’ve made it less so, but altering concurrency-related code seems like a risk we’re better off avoiding.

I can’t see anything wrong with your reasoning, but this sort of stuff is prone to “surprises”. Will this palpably improve speed? If so, then I’m OK with it. If not, IMO we should retain the present logic.

FGasper (approved):
LGTM % the last few small notes I’ve made.

internal/verifier/change_stream.go
tdq45gj: Technically none of them need to be exported since they're all in the `verifier` package. Changed to private.

internal/verifier/migration_verifier_test.go
tdq45gj: Looks like it's more likely to race if inserts happen after generation start because StartNextGeneration isn't blocked on the next generation. I've reverted the changes for this test so that change events occur in the previous generation and tasks should appear in the next generation.

PR #54, opened 2024-11-25
Set `showExpandedEvents` in change streams for server v6+.

FGasper:
Since migration-verifier forbids DDL events during verification, it should set `showExpandedEvents` so that it can fail on events like `create` or `shardCollection`.


internal/verifier/change_stream_test.go
```diff
@@ -291,3 +307,36 @@ func (suite *IntegrationTestSuite) TestEventBeforeWritesOff() {
...
+
+	db := suite.srcMongoClient.Database(suite.DBNameForTest())
+	coll := db.Collection("mycoll")
+	suite.Require().NoError(
+		db.CreateCollection(ctx, coll.Name()),
+	)
+
+	suite.Require().NoError(verifier.WritesOff(ctx))
+
+	err = verifierRunner.Await()
```
tdq45gj, line 336: I don't think a change stream error will be returned until writesOff

FGasper: It’s actually indeterminate, it seems: the error may happen before writesOff, or it may happen when shutting down the change stream.

PR #56, opened 2024-11-25
Make CheckRunner methods check to see if the verifier failed.

FGasper:
This prevents the test from hanging if the verifier fails for an unforeseen reason.


tdq45gj (approved):
LGTM

PR #57, opened 2024-11-25
REP-5318 Hide `updateDescription` in update events.

FGasper:
This prevents one means of triggering a BSONObjectTooLarge error. Such errors could still happen from over-large `_id` fields, but users are unlikely to have them.

Found while running migration-verifier’s change stream against mongosync’s nondeterministic tests.


tdq45gj (approved):
LGTM

PR #58, opened 2024-11-26
Read documents in parallel

FGasper:
This tweaks the document comparison from REP-5230 (PR #34) so that, instead of reading source and destination documents in series, they are read in parallel. This should improve comparison speed.

This copies syncmap and syncslice from mongosync.


tdq45gj (approved):
LGTM

PR #61, opened 2024-11-27
REP-5330 Fix deep index comparison.

FGasper:
This adds a $literal around the documents inserted into the $documents aggregation to allow for number-type discrepancies in index specifications. See the test case added here for why this is needed.


tdq45gj (approved):
LGTM!

PR #65, opened 2024-11-28
Add change stream lag to periodic progress reports.

FGasper:
This adds a report of change stream lag to the logs’ progress reports.


FGasper:
Recreated as #74.

PR #66, opened 2024-11-29
REP-5329 Teach retryer to run +1 callback in parallel.

FGasper:
We need retry logic around `find` cursor iteration, but the retryer’s IterationSuccess() method won’t work properly with how we compare partitions document-by-document. If, for example, the source hangs but we call IterationSuccess() on every destination success, we’ll never time out the retries. But we need IterationSuccess() to work or else there are no retries for partitions that take a long time to read.

This changeset fixes that by rewriting the retryer to take multiple callbacks. Each callback now takes a context that the retryer uses to stop the function whenever another thread has failed.

Since migration-verifier doesn’t handle DDL events, there’s no point to handling collection UUIDs. Thus, the initial commits in this PR resolve longstanding technical debt.

This also synchronizes migration-verifier’s internal list of transient errors with mongosync’s.

### DIFF
--- a/internal/verifier/compare.go
+++ b/internal/verifier/compare.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"time"
 
+	"github.com/10gen/migration-verifier/internal/retry"
 	"github.com/10gen/migration-verifier/internal/types"
 	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson"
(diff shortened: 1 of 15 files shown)


internal/verifier/compare.go
```diff
@@ -308,16 +315,26 @@ func (verifier *Verifier) getFetcherChannels(
...
+	cursor *mongo.Cursor,
+	writer chan<- bson.Raw,
+) error {
 	for cursor.Next(ctx) {
+		state.NoteSuccess()
 		writer <- slices.Clone(cursor.Current)
 	}
 
+	if cursor.Err() == nil {
+		state.NoteSuccess()
```
tdq45gj, line 335: Looks like the retryer refreshes `lastResetTime` once the error group finishes. Is it redundant to call `NoteSuccess` before the callback function returns?

tdq45gj (approved):
LGTM % one small question

internal/verifier/compare.go
FGasper: Probably. Removed.

edobranov (approved):
LGTM

PR #69, opened 2024-12-03
REP-5329 Fix retry on document reads.

FGasper:
Commit 31cc703daf was broken because it didn’t actually recreate the callbacks as needed to redo the reads.

(Unfortunately there’s little good way to test this directly, but mongosync’s CI testing caught it.)

This changeset fixes that by adding a callback to the retryer that runs at the start of each iteration.

### DIFF
--- a/internal/retry/retry.go
+++ b/internal/retry/retry.go
@@ -44,9 +44,9 @@ type RetryCallback = func(context.Context, *FuncInfo) error
 // This returns an error if the duration limit is reached, or if f() returns a
 // non-transient error.
 func (r *Retryer) Run(
-	ctx context.Context, logger *logger.Logger, f ...RetryCallback,
+	ctx context.Context, logger *logger.Logger, funcs ...RetryCallback,
 ) error {
-	return r.runRetryLoop(ctx, logger, f)
+	return r.runRetryLoop(ctx, logger, funcs)
 }
 
 // runRetryLoop contains the core logic for the retry loops.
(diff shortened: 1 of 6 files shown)


internal/retry/retry.go
```diff
@@ -74,10 +74,23 @@ func (r *Retryer) runRetryLoop(
...
+		}
+
 		eg, egCtx := errgroup.WithContext(ctx)
 		for i, curFunc := range funcs {
+			if curFunc == nil {
+				panic("curFunc should be non-nil")
+			}
 
 			eg.Go(func() error {
+				if curFunc == nil {
```
tdq45gj, line 88: Why do we check curFunc to be non-nil both twice?

internal/verifier/compare.go
```diff
@@ -25,31 +25,40 @@ func (verifier *Verifier) FetchAndCompareDocuments(
...
-			results, docCount, byteCount, err = verifier.compareDocsFromChannels(
-				ctx,
-				task,
-				srcChannel,
-				dstChannel,
-			)
+	err := retryer.
+		WithBefore(func() {
+			srcChannel, dstChannel, readSrcCallback, readDstCallback = verifier.getFetcherChannelsAndCallbacks(task)
+		}).
```
tdq45gj, line 40: I'm trying to understand why do we need before func. Is it the same as calling the function at the start of the retried function except that the error from it won't be retried?

tdq45gj (approved):
LGTM % two questions.

internal/retry/retry.go
FGasper: This 2nd check is probably redundant. I was just trying to suss out why I was seeing segfaults, and I figured I might as well leave the panics in.

internal/verifier/compare.go
FGasper: The trick is that there are _multiple_ retried funcs. Having a “before” func guarantees execution of one piece before all the others … which is crucial here because the callbacks & channels have to get recreated every time.

I’ll clarify this in docs.

PR #71, opened 2024-12-04
REP-5317 Ignore mongosync DBs on destination.

FGasper:
REVIEWER NOTE: The first couple commits here are general cleanup. This PR will be merged via rebase.

-----

To make migration-verifier’s change stream work with mongosync CI we need to ignore mongosync’s internal databases, which mongosync will manipulate alongside migration-verifier.

This changeset adjusts the change stream logic accordingly.


tdq45gj (approved):
LGTM, thanks for this fix!

PR #72, opened 2024-12-04
Add a few additional data points to logs.

FGasper:
This adds:
- “Processing document comparison task” now mentions a) the namespace, and b) either the range’s min & max (gen0/check) or the # of documents (gen1+/recheck).
- “Finished document comparison task” now mentions the namespace.
- Logs for creation of recheck tasks now give the task ID.
- A new “Scheduled documents for recheck in the new generation” log that gives the total # and size of rechecks.

These data points would have been nice conveniences when reviewing logs for InfoMedia’s recent migration.


tdq45gj (approved):
LGTM

PR #74, opened 2024-12-04
Add change stream lag to periodic progress reports.

FGasper:
This adds a report of change stream lags to the logs’ progress reports.

### DIFF
--- a/internal/verifier/summary.go
+++ b/internal/verifier/summary.go
@@ -399,43 +399,59 @@ func (verifier *Verifier) printChangeEventStatistics(builder *strings.Builder) {
 
 	builder.WriteString(fmt.Sprintf("\nChange events this generation: %s\n", eventsDescr))
 
-	if totalEvents == 0 {
-		return
+	if totalEvents > 0 {
+
+		reverseSortedNamespaces := maps.Keys(nsTotals)
+		sort.Slice(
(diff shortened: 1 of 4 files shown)


internal/verifier/summary.go
```diff
@@ -399,43 +399,59 @@ func (verifier *Verifier) printChangeEventStatistics(builder *strings.Builder) {
...
+			fmt.Sprintf("\nSource change stream lag: %s\n", reportutils.DurationToHMS(srcLag)),
 		)
 	}
 
-	builder.WriteString("\nMost frequently-changing namespaces:\n")
-	table.Render()
+	dstLag, hasDstLag := verifier.dstChangeStreamReader.GetLag().Get()
+	if hasDstLag {
+		if !hasSrcLag {
+			builder.WriteString("\n")
```
tdq45gj, line 449: It seems that src and dst lags are adding the new line differently. Not sure if it's intentional.

tdq45gj (approved):
LGTM

internal/verifier/summary.go
FGasper: There should always be a single newline in front of the lag-report lines. So if there was a source lag then there’s no need for one before the dst lag, but if there wasn’t a source lag then the dst lag has to have one.

PR #76, opened 2024-12-05
REP-5358 Revamp retryer.

FGasper:
This improves the retryer in several ways that should aid in debugging such failures as REP-5358:

- There are now descriptions for each callback and (optionally) the loop itself. Those descriptions go in the relevant log messages.
- NoteSuccess() also takes a description that goes into logs.
- Retried failures are now noted at info level rather than debug.
- The retryer now notes success of a retried operation.
- Long-running operations that go without a success for >1min will be noted in the log as a warning.
- The “retry randomly” functionality is removed. (None of the existing tests use it.)

The above requires a revamp of the retryer interface, so all callers into it are rewritten.

### DIFF
--- a/internal/partitions/partitions.go
+++ b/internal/partitions/partitions.go
@@ -119,7 +119,6 @@ const (
 func PartitionCollectionWithSize(
 	ctx context.Context,
 	uuidEntry *uuidutil.NamespaceAndUUID,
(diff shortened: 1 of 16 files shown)


tdq45gj (approved):
LGTM

PR #78, opened 2024-12-09
REP-5358 Add additional retryer updates.

FGasper:
PR #76 missed some changes due to differences between repositories. This changeset restores those:
- NoteSuccess() now requires a description of what succeeded, which will go into the log.
- The retryer logs context cancellation errors at debug level.
- A few more data points go into the logs now.


tdq45gj (approved):
LGTM

PR #79, opened 2024-12-10
Fix change stream lag calculation.

FGasper:
The previous logic was backwards, which caused uint overflow.

This adds a test that usually passes without this change but should occasionally fail. (It should always pass, though, with this change.)

### DIFF
--- a/internal/verifier/change_stream.go
+++ b/internal/verifier/change_stream.go
@@ -318,7 +318,7 @@ func (csr *ChangeStreamReader) readAndHandleOneChangeEventBatch(
 	var curTs primitive.Timestamp
 	curTs, err := extractTimestampFromResumeToken(cs.ResumeToken())
 	if err == nil {
-		lagSecs := int32(sess.OperationTime().T) - int32(curTs.T)
+		lagSecs := int64(sess.OperationTime().T) - int64(curTs.T)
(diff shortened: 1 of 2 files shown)


internal/verifier/change_stream.go
```diff
@@ -318,7 +318,7 @@ func (csr *ChangeStreamReader) readAndHandleOneChangeEventBatch(
 	var curTs primitive.Timestamp
 	curTs, err := extractTimestampFromResumeToken(cs.ResumeToken())
 	if err == nil {
-		lagSecs := curTs.T - sess.OperationTime().T
+		lagSecs := int32(sess.OperationTime().T) - int32(curTs.T)
```
tdq45gj, line 321: This seems to be prone to overflow. Can we convert `T` (uint32) to int64?

tdq45gj (changes requested):
The PR comment says it adds a test. Are we still adding it?

internal/verifier/change_stream.go
FGasper: yes, good call

FGasper:
@tdq45gj I added the test. (Somehow I omitted it inadvertently earlier.)

tdq45gj (approved):
LGTM, thanks!

PR #85, opened 2025-01-21
REP-5550 Add resumability for recheck generations

FGasper:
Previously migration-verifier did not persist its generation, so if it crashed in a generation after generation 0 it would still restart at generation 0.

This could cause any enqueued rechecks not to happen until the restarted verifier reached the prior verifier run’s generation. If the user called `writesOff` before then, though, the rechecks would never happen. Thus, migration-verifier could neglect important rechecks.

This fixes that by persisting migration-verifier’s generation at the start of each generation. Thus, any tasks that need to be rerun, or enqueued rechecks that need to be processed into checks for the next generation, will be replayed in the event of a crash.

### DIFF
--- a/internal/verifier/check.go
+++ b/internal/verifier/check.go
@@ -196,6 +196,20 @@ func (verifier *Verifier) CheckDriver(ctx context.Context, filter map[string]any
 		if err != nil {
 			return err
 		}
+	} else {
+		genOpt, err := verifier.readGeneration(ctx)
+		if err != nil {
+			return err
+		}
+
+		if gen, has := genOpt.Get(); has {
+			verifier.generation = gen
+			verifier.logger.Info().
+				Int("generation", verifier.generation).
+				Msg("Resuming in-progress verification.")
+		} else {
+			verifier.logger.Info().Msg("Starting new verification.")
+		}
 	}
 	err = retry.New().WithCallback(
 		func(ctx context.Context, _ *retry.FuncInfo) error {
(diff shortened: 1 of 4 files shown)


internal/verifier/check.go
```diff
@@ -196,7 +195,27 @@ func (verifier *Verifier) CheckDriver(ctx context.Context, filter map[string]any
 		if err != nil {
```
tdq45gj, line 195: now that we're moving the unlock further down, we probably need to unlock before returning

internal/verifier/check.go
```diff
@@ -196,7 +195,27 @@ func (verifier *Verifier) CheckDriver(ctx context.Context, filter map[string]any
...
 			return err
 		}
+	} else {
+		genOpt, err := verifier.readGeneration(ctx)
+		if err != nil {
+			return err
```
tdq45gj, line 201: and also unlock here before returning

tdq45gj (changes requested):
LGTM % locking

internal/verifier/generation.go
```diff
@@ -0,0 +1,63 @@
...
+	generationCollName  = "generation"
+	generationFieldName = "generation"
+)
+
+func (v *Verifier) persistGenerationWhileLocked(ctx context.Context) error {
+	generation, _ := v.getGenerationWhileLocked()
+
+	db := v.verificationDatabase()
+
+	result, err := db.Collection(generationCollName).ReplaceOne(
```
jtcovan, line 24: I'm not super familiar with the migration verifier code. I just want to check that we won't have any races where multiple verifiers could be updating this value and we end up persisting an incorrect generation? The MV runs serial generations so this shouldn't happen, right? And you can't run multiple MV simultaneously? Or if you can, they have different db names?

internal/verifier/check.go
jtcovan: Will the defer handle this clean up?

FGasper: Actually, no, the defer handle would block, I think … which would be fun.

Fixed.

internal/verifier/check.go
FGasper: done

internal/verifier/generation.go
FGasper: No 2 MV instances can share the same metadata, so there should be no such races.

tdq45gj (approved):
LGTM, thanks!

jtcovan (approved):
LGTM

PR #86, opened 2025-01-23
REP-5470 Fix hang when handler thread hits an error.

FGasper:
This fixes a hang that previously happened if migration-verifier received an event that it couldn’t handle.

Specific changes:
- The event optype and namespace presence are checked right after decoding. Previously this happened in the handler, which was the immediate cause of the hang that this changeset fixes. Doing this check makes the failure faster and more prominent.
- If the handler fails, it stores its error, and the reader checks for that error.
- Many debug-level logs are added.
- Context cancellations/timeouts are reported with their Cause.
- WritesOff() is rewritten to make it easier to reason about the mutex handling’s correctness.
- The generation-end error check no longer mistakenly assumes that, if the worker-controller thread finished without error, then all is well.

This also removes a flapping assertion in `TestStartAtTimeNoChanges`.

### DIFF
--- a/internal/verifier/change_stream.go
+++ b/internal/verifier/change_stream.go
@@ -78,7 +78,8 @@ type ChangeStreamReader struct {
 	changeStreamRunning  bool
 	changeEventBatchChan chan []ParsedEvent
 	writesOffTs          *util.Eventual[primitive.Timestamp]
-	error                *util.Eventual[error]
+	readerError          *util.Eventual[error]
+	handlerError         *util.Eventual[error]
 	doneChan             chan struct{}
 
 	startAtTs *primitive.Timestamp
@@ -97,7 +98,8 @@ func (verifier *Verifier) initializeChangeStreamReaders() {
 		changeStreamRunning:  false,
 		changeEventBatchChan: make(chan []ParsedEvent),
 		writesOffTs:          util.NewEventual[primitive.Timestamp](),
-		error:                util.NewEventual[error](),
+		readerError:          util.NewEventual[error](),
+		handlerError:         util.NewEventual[error](),
 		doneChan:             make(chan struct{}),
 		lag:                  msync.NewTypedAtomic(option.None[time.Duration]()),
 	}
@@ -111,7 +113,8 @@ func (verifier *Verifier) initializeChangeStreamReaders() {
 		changeStreamRunning:  false,
 		changeEventBatchChan: make(chan []ParsedEvent),
 		writesOffTs:          util.NewEventual[primitive.Timestamp](),
-		error:                util.NewEventual[error](),
+		readerError:          util.NewEventual[error](),
+		handlerError:         util.NewEventual[error](),
 		doneChan:             make(chan struct{}),
 		lag:                  msync.NewTypedAtomic(option.None[time.Duration]()),
 	}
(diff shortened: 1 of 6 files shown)


internal/verifier/check.go
```diff
@@ -313,12 +313,21 @@ func (verifier *Verifier) CheckDriver(ctx context.Context, filter map[string]any
...
-				return err
-			}
-			if err = verifier.waitForChangeStream(ctx, verifier.dstChangeStreamReader); err != nil {
-				return err
+
+			for _, csr := range mslices.Of(verifier.srcChangeStreamReader, verifier.dstChangeStreamReader) {
+				if err = verifier.waitForChangeStream(ctx, csr); err != nil {
+					return errors.Wrapf(
+						err,
+						"failed to close %s",
```
tdq45gj, line 321: How is the error related to closing a change stream reader?

internal/verifier/change_stream_test.go
```diff
@@ -303,6 +303,39 @@ func (suite *IntegrationTestSuite) fetchVerifierRechecks(ctx context.Context, ve
...
+	verifier := suite.BuildVerifier()
+
+	suite.Require().NoError(
+		db.CreateCollection(ctx, "mycoll"),
+	)
+
+	verifier.SetSrcNamespaces([]string{db.Name() + ".mycoll"})
+	verifier.SetDstNamespaces([]string{db.Name() + ".mycoll"})
+	verifier.SetNamespaceMap()
+	// verifier.SetVerifyAll(true)
```
tdq45gj, line 323: why is this line commented out?

tdq45gj (changes requested):
LGTM % two small things.

internal/verifier/change_stream.go
```diff
@@ -103,31 +113,55 @@ func (verifier *Verifier) initializeChangeStreamReaders() {
...
-			return ctx.Err()
+			err = util.WrapCtxErrWithCause(ctx)
+
+			verifier.logger.Debug().
+				Err(err).
+				Stringer("changeStreamReader", reader).
+				Msg("Closing change event handler.")
+
 		case batch, more := <-reader.changeEventBatchChan:
 			if !more {
```
autarch, line 140: This isn't related to this PR, but it occurs to me that this may be a bug. I'm pretty sure that `batch` can be non-empty even when `more` is false.

autarch (changes requested):
This is a fairly large and complex diff for a codebase I don't understand very well. It _looks_ okay, but there's a lot going on here. It seems like there's a lot more changes than the bare minimum needed to prevent the hang. Did you do some significant refactoring as well?

Would it be possible to split this up into multiple PRs?

internal/verifier/change_stream.go
FGasper: I think that’s incorrect: if `more` is false, `batch` is always the zero value.

[Go by Example](https://gobyexample.com/closing-channels)

internal/verifier/change_stream.go
```diff
@@ -103,31 +113,55 @@ func (verifier *Verifier) initializeChangeStreamReaders() {
...
+				Msg("Handling change event batch.")
+
+			err = errors.Wrap(
+				verifier.HandleChangeStreamEvents(ctx, batch, reader.readerType),
+				"failed to handle change stream events",
+			)
 		}
 	}
+
+	// This will prevent the reader from hanging.
```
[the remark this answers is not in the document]
FGasper: done

internal/verifier/change_stream_test.go
FGasper: cruft. REmoving.

internal/verifier/check.go
FGasper: I’ll reword.

FGasper:
> It seems like there's a lot more changes than the bare minimum needed to prevent the hang.

I left a lot of debug logging in because they seem potentially useful and unlikely to pollute logs (excessively, anyhow).

> Did you do some significant refactoring as well?

No, actually. There’s some renaming (e.g., `error` to `readerError`), and maybe what you’re seeing is the moving of the optype validation to sooner in the change stream workflow, but I wouldn’t call anything here a significant refactor.

> Would it be possible to split this up into multiple PRs?

It seems a mite excessive for a <1k-line PR, most of whose changes are logging, but of course I wrote the changes. :)

I can go back over this and make discrete commits.

FGasper:
@autarch I’ve separated this into discrete commits.

tdq45gj (approved):
LGTM

autarch:
> > Would it be possible to split this up into multiple PRs?
> 
> It seems a mite excessive for a <1k-line PR, most of whose changes are logging, but of course I wrote the changes. :)
> 
> I can go back over this and make discrete commits.

I think the reason I'm struggling with this is that I don't particularly understand the state of the code _before_ the PR. So this seems like a fairly large change to review in that context.

I think discrete commits would make this a lot easier for me to review. I did take a look at the commit history but it didn't look amenable to a commit-by-commit review, which is part of why I asked if this could be broken up differently.

internal/verifier/change_stream.go
autarch: Ah, I misunderstood some writing about channels by Dave Cheney, I think.

autarch (approved):
LGTM!

PR #87, opened 2025-02-11
Update README

tdq45gj:
An update to README per @ajayvijayakumar123's request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,11 @@ _If verifying a migration done via [mongosync](https://www.mongodb.com/docs/clus
 
 _Assumes no port set, default port for operation webserver is 27020_
 
+# Recommendations
+
+Users should use a different cluster for storing migration verification metadata. If the user must use the destination cluster,
+they need to either disable mongosync's write blocking on the destination or provide a `--metaURI` with a database user that has the `bypassWriteBlockingMode`
+permission.
 
 ## Launch the Verifier Binary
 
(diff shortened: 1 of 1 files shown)


ajayvijayakumar123 (approved):
LGTM!

README.md
```diff
@@ -14,6 +14,11 @@ _If verifying a migration done via [mongosync](https://www.mongodb.com/docs/clus
 
 _Assumes no port set, default port for operation webserver is 27020_
 
+# Recommendations
+
+Users should use a different cluster for storing migration verification metadata. If the user must use the destination cluster,
```
FGasper, line 19: We should clarify _why_.

FGasper (changes requested):

README.md
FGasper: Also, the user may _be_ the reader of this document.

FGasper: Suggested:
```
# Verifier Metadata Considerations

migration-verifier needs a database to store its state. This database SHOULD be on its own cluster.

The verifier _can_ instead store its metadata on the destination cluster. This can severely degrade performance, though. It also requires either disabling mongosync’s destination write blocking or giving the `bypassWriteBlockingMode` to the verifier’s `--metaURI` user.
```

ajayvijayakumar123: yup fair point. Also, I asked Jian to put up this PR because I'm still lacking push permissions for some reason (and I didn't want to deal with all the permission stuff yet for a readme update 😅 )

FGasper (approved):
LGTM. Thanks!

PR #88, opened 2025-02-14
Update logging & documentation for binding to port 0

FGasper:
It’s useful in testing to bind to an ephemeral port so as to avoid port conflicts. This changeset improves logging & documentation around this.

Documentation also explicitly now states that migration-verifier binds to all available network interfaces.


README.md
```diff
@@ -53,9 +53,11 @@ migration-verifier … --srcNamespace foo.bar --srcNamespace foo.baz --dstNamesp
 
 will result in the mapping `foo.bar <-> foo.bar1, foo.baz <-> bar.bar2`
 
-By default, the verifier will read from the primary node.  This can be changed with option “`--readPreference <preference>`” where `<preference>` can be “`primary`” (same as default), “`secondary`”, “`primaryPreferred`”, “`secondaryPreferred`”, or “`nearest`”.  
+By default, the verifier will read from the primary node.  This can be changed with option “`--readPreference <preference>`” where `<preference>` can be “`primary`” (same as default), “`secondary`”, “`primaryPreferred`”, “`secondaryPreferred`”, or “`nearest`”.
```
tdq45gj, line 56: Can we add a warning about the risk of using `--readPreference secondary` with sharded clusters and link to the documentation page: https://www.mongodb.com/docs/manual/core/read-preference/#behavior?

tdq45gj (changes requested):
It looks good. This might be unrelated to the main purpose of this PR, but I left one comment about how we are documenting the `readPreference` option.

README.md
FGasper: Yeah this seems unrelated, and TBH I’m not sure we should even _allow_ setting the read pref if this is a concern. Or, maybe we should disallow setting the read pref against sharded clusters.

I’ll open a separate ticket to discuss.

tdq45gj (approved):
LGTM, thanks!

PR #90, opened 2025-02-24
Distinguish src vs. dst events; compute lag even without events.

FGasper:


internal/verifier/summary.go
```diff
@@ -381,76 +381,81 @@ func (verifier *Verifier) printMismatchInvestigationNotes(strBuilder *strings.Bu
...
+	}{
+		{"Source", verifier.srcEventRecorder, verifier.srcChangeStreamReader},
+		{"Destination", verifier.dstEventRecorder, verifier.dstChangeStreamReader},
+	} {
+		nsStats := cluster.eventRecorder.Read()
+
+		activeNamespacesCount := len(nsStats)
+
+		totalEvents := 0
+		nsTotals := map[string]int{}
```
tdq45gj, line 399: nit: the map could be initialized with size `len(nsStats)`.

tdq45gj (approved):
LGTM

internal/verifier/summary.go
FGasper: I’d take this suggestion were this new code, but it’s just moved/indented from before, so I’d rather leave it.

PR #92, opened 2025-03-01
REP-5621 migration-verifier: Report field-order-sensitive mismatch details

khodakovski:
Report mismatch details in the case of field-order-sensitive verification:
1. Report mismatched fields if the field content is different.
2. Report different field order if the field content matches but the field order is different.

### DIFF
--- a/internal/verifier/migration_verifier_test.go
+++ b/internal/verifier/migration_verifier_test.go
@@ -22,7 +22,6 @@ import (
 	"github.com/10gen/migration-verifier/mslices"
 	"github.com/cespare/permute/v2"
 	"github.com/rs/zerolog"
-	"github.com/samber/lo"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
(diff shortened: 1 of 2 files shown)


internal/verifier/migration_verifier_test.go
```diff
@@ -99,28 +98,21 @@ func (suite *IntegrationTestSuite) TestVerifierFetchDocuments() {
...
 
 	verifier.globalFilter = map[string]any{"num": map[string]any{"$lt": 100}}
 	results, docCount, byteCount, err = verifier.FetchAndCompareDocuments(ctx, task)
 	suite.Require().NoError(err)
 	suite.Assert().EqualValues(1, docCount, "should find source docs")
-	suite.Assert().NotZero(byteCount, "should tally docs’ size")
+	suite.Assert().NotZero(byteCount, "should tally docs' size")
 	suite.Require().Len(results, 1)
-	suite.Assert().Equal(Mismatch, results[0].Details)
+	suite.Assert().Regexp(regexp.MustCompile("^"+Mismatch), results[0].Details, "mismatch expeceted")
```
autarch, line 115: In general, it's not safe to just embed a string directly in a regex. The string could contain regex metacharacters like `.` or `*`. You need to use `\Q` to embed the string:

```
fmt.Sprintf(`^\Q%s\E`, Mismatch)
```

internal/verifier/migration_verifier_test.go
```diff
@@ -135,9 +127,9 @@ func (suite *IntegrationTestSuite) TestVerifierFetchDocuments() {
...
```
autarch, line 132: same issue here

internal/verifier/migration_verifier_test.go
```diff
@@ -99,28 +98,21 @@ func (suite *IntegrationTestSuite) TestVerifierFetchDocuments() {
...
-		lo.Map(
-			results,
-			func(result VerificationResult, _ int) any {
-				return result.Details
-			},
-		),
-		"details as expected",
-	)
+	for _, res := range results {
+		suite.Assert().Regexp(regexp.MustCompile("^"+Mismatch), res.Details, "details as expected")
```
autarch, line 104: same regex issue here

internal/verifier/migration_verifier_test.go
khodakovski: I will fix it in a separate PR since this pattern is everywhere in this code base.

PR #96, opened 2025-03-12
Add events per second to log, and use same “now” in progress logs.

FGasper:
This adds a simple events-per-second computation to the progress logs, which will provide a useful at-a-glance metric of how active the change streams are. It also changes the logic a bit to compute all time deltas from the same “now”.


tdq45gj (approved):
LGTM

PR #97, opened 2025-03-21
Ignore/log `modify` events on the destination.

FGasper:
When mongosync migrates capped collections it sets the collection limits during commit, which causes the destination to emit `modify` events in its change stream. These have caused migration-verifier to fail because of the “unrecognized” event type.

This changeset alters that logic so that migration-verifier will log `modify` events from the destination and ignore them rather than treating them as fatal.

Additionally, this updates the unknown-event error to include the change event in its bson.Raw form rather than a struct. This ensures that we show the entire event rather than just those parts that we parse.

### DIFF
--- a/internal/verifier/change_stream.go
+++ b/internal/verifier/change_stream.go
@@ -360,19 +360,21 @@ func (csr *ChangeStreamReader) readAndHandleOneChangeEventBatch(
 			// We expect modify events on the destination as part of finalizing
 			// a migration. For example, mongosync enables indexes’ uniqueness
 			// constraints and sets capped collection sizes.
-			if opType == modifyEventType && csr.onModifyEvent == onModifyEventIgnore {
-				csr.logger.Info().
-					Stringer("changeStream", csr).
-					Interface("event", cs.Current).
-					Msg("This event is probably internal to the migration. Ignoring.")
+			/*
+				if opType == modifyEventType && csr.onModifyEvent == onModifyEventIgnore {
+					csr.logger.Info().
+						Stringer("changeStream", csr).
+						Interface("event", cs.Current).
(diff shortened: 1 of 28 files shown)


internal/verifier/change_stream.go
```diff
@@ -343,8 +355,25 @@ func (csr *ChangeStreamReader) readAndHandleOneChangeEventBatch(
...
 
-		if !supportedEventOpTypes.Contains(changeEventBatch[eventsRead].OpType) {
-			return UnknownEventError{Event: &changeEventBatch[eventsRead]}
+		opType := changeEventBatch[eventsRead].OpType
+		if !supportedEventOpTypes.Contains(opType) {
+
+			// We expect modify events on the destination as part of finalizing
+			// a migration. For example, mongosync enables indexes’ uniqueness
+			// constraints and sets capped collection sizes.
+			if opType == modifyEventType && csr.onModifyEvent == onModifyEventIgnore {
```
tdq45gj, line 364: A random thought: there are some other DDL events that could be caused by mongosync on the destination, such as create and rename. Should we ignore those as well?

tdq45gj (approved):
LGTM. I left one question. I don't think we have heard complaints about it, and I might be mistaking.

internal/verifier/change_stream.go
FGasper: MV doesn’t allow DDL on the source during a migration, so I can’t think of any reason why there’s be a create during a migration.

I thank we can just allow more events as requests arrive.

tdq45gj: Mongosync would still create on the destination at the start of a migration. But probably not a problem because we recommend starting MV after Collection Copy.

PR #99, opened 2025-03-27
Migrate to contextplus for more informative “context canceled” errors.

FGasper:
```
2025-03-27T08:52:49-04:00 DBG Non-retryable error occurred. error="context canceled" description=["finding next task to do in generation 0"] error code=0
```
^^ The above doesn’t tell us _why_ the context was canceled.

This changeset causes all contexts’ Err() to show their cancellation causes, e.g:
```
2025-03-27T08:52:49-04:00 DBG Non-retryable error occurred. error="context canceled: generation 0 finished" description=["finding next task to do in generation 0"] error code=0
```

Note the addition of `generation 0 finished`, which lets us know why the context was canceled.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,7 @@
 linters:
   enable:
     - errcheck
+    - forbidigo
     - gosimple
     - govet
     - ineffassign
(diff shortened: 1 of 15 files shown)


tdq45gj (approved):
I think this is good. LGTM!

PR #104, opened 2025-04-08
REP-5904 Fix BSON size calculation when creating recheck tasks

FGasper:
When creating tasks from enqueued rechecks, migration-verifier groups multiple rechecks together per task. The status quo logic added the length of all document `_id`s and set a limit at 12 MiB.

This logic, though, only included the length of the `_id` _values_, not their type bytes nor their array indexes. (NB: BSON arrays are encoded identically to BSON embedded documents with keys like `1`, `2`, etc.) Thus, if the number of documents to recheck is high enough, and the individual `_id`s are small enough, it’s possible to yield an over-large BSON document.

This changeset fixes that by making the recheck logic determine the real BSON size of the array of `_id`s.

To add a suitable test for this it was necessary to insert millions of documents into the recheck queue. That revealed a performance problem with our current recheck insertion logic: it uses upsert, which is markedly slower than a plain insert. Thus, tests that need to enqueue so many rechecks will time out. To fix that, this changeset switches to bulk _inserts_ instead of upserts, with `ordered: false` and additional logic to tolerate bulk-write errors.


FGasper:
## Review Notes

Please review this PR’s commits individually. Two are general “housekeeping”, and two are the actual changes. I will merge them as-is rather than squashing.

internal/verifier/recheck.go
```diff
@@ -137,6 +130,12 @@ func (verifier *Verifier) insertRecheckDocs(
 						options.BulkWrite().SetOrdered(false),
 					)
 
+					err = util.TolerateSimpleDupeKeyInBulk(
```
tdq45gj, line 136: Where does the first recheck document come from? It makes sense to ignore the error if it only comes from the last retry or a previous MV run. Could we add a comment here to explain?

tdq45gj (commented):
This change looks good. Left a question that might worth explaining.

internal/verifier/recheck.go
FGasper: I’ve added this comment:
```
+                                       // We expect duplicate-key errors from the above because:
+                                       //
+                                       // a) The same document can change multiple times per generation.
+                                       // b) The source’s changes also happen on the destination, and both
+                                       //    change streams populate this recheck queue.
+                                       //
+                                       // Because of that, we ignore duplicate-key errors. We do *not*, though,
+                                       // ignore other errors that may accompany duplicate-key errors in the
+                                       // same server response.
```

internal/verifier/recheck_test.go
```diff
@@ -200,6 +206,68 @@ func (suite *IntegrationTestSuite) TestRecheckResumability_Mismatch() {
...
+		suite.T().Name(), "testColl",
+		lo.ToAnySlice(ids),
+		lo.RepeatBy(docsCount, func(_ int) int { return 16 }),
+	)
+	suite.Require().NoError(err, "should insert the second time")
+}
+
+func (suite *IntegrationTestSuite) TestManyManyRechecks() {
+	if len(os.Getenv("CI")) > 0 {
+		suite.T().Skip("Skipping this test in CI. (It causes GitHub Action to self-terminate.)")
```
tdq45gj, line 240: We should run the tests on Evergreen. Could we file a ticket?

internal/verifier/recheck.go
tdq45gj: Thanks for the explanation

tdq45gj (approved):
LGTM, thanks!

internal/verifier/recheck_test.go
FGasper: Filed (REP-5907).

FWIW I actually think the benefit of Evergreen is minimal here since this is a F/OSS project. But it would be good to run this test, at least. I wonder if there’s a GHA setting we could tweak.

internal/util/bson.go
```diff
@@ -0,0 +1,37 @@
...
+	count     int
+}
+
+// Add adds a new element.
+func (b *BSONArraySizer) Add(el bson.RawValue) {
+	newElIndex := strconv.Itoa(b.count)
+
+	// Each element contains:
+	//	- a type byte
+	//  - the stringified key
```
janetio, line 22: Can we reorder so that NUL byte is second in the list? Just so its easier to match up to the addition below.

internal/util/bson.go
```diff
@@ -0,0 +1,37 @@
...
+	//  - a NUL byte
+	//  - the encoded BSON value
+	//
+	// We thus augment b.size accordingly:
+	b.elemsSize += 2 + len(newElIndex) + len(el.Value)
+
+	b.count++
+}
+
+// Len() returns the encoded BSON array’s length in bytes.
```
janetio, line 32:
```suggestion
// Len returns the encoded BSON array’s length in bytes.
```
To match the definition for `Add`

internal/util/bson_test.go
```diff
@@ -0,0 +1,44 @@
+package util
+
+import (
+	"github.com/10gen/migration-verifier/mbson"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/bson/primitive"
+)
+
+func (s *UnitTestSuite) TestBSONArrayBuilder() {
```
janetio, line 9:
```suggestion
func (s *UnitTestSuite) TestBSONArraySizer() {
```

internal/util/bson.go
FGasper: The list is ordered as BSON does it.

FGasper: I’ll split it out into:
```
b.elemsSize += 1 + len(newElIndex) + 1 + len(el.Value)
```

internal/verifier/recheck_test.go
```diff
@@ -200,6 +206,68 @@ func (suite *IntegrationTestSuite) TestRecheckResumability_Mismatch() {
...
+	err := insertRecheckDocs(
+		ctx,
+		verifier,
+		suite.T().Name(), "testColl",
+		lo.ToAnySlice(ids),
+		lo.RepeatBy(docsCount, func(_ int) int { return 16 }),
+	)
+	suite.Require().NoError(err, "should insert the first time")
+
+	err = insertRecheckDocs(
```
janetio, line 228: Is there a reason we don't call verifier.insert...?

internal/verifier/recheck_test.go
janetio: Why does this cause the GHA to self-terminate?

internal/verifier/recheck_test.go
```diff
@@ -200,6 +206,68 @@ func (suite *IntegrationTestSuite) TestRecheckResumability_Mismatch() {
...
+		ctx,
+		verifier,
+		suite.T().Name(), "testColl",
+		lo.ToAnySlice(ids),
+		lo.RepeatBy(docsCount, func(_ int) int { return 16 }),
+	)
+	suite.Require().NoError(err, "should insert the second time")
+}
+
+func (suite *IntegrationTestSuite) TestManyManyRechecks() {
```
janetio, line 238: Would the ids created by `lo.Range()` in this test encounter the original overly-large bson doc problem that existed before this PR?

FGasper: It did in testing, yes.

The issue is that the old logic, which just added up the lengths of the bson.RawValue.Value buffers, didn’t account for the extra stuff in a BSON document. If you have a million doc IDs, then each byte used to encode an ID consumes 1 MiB in toto. So to make it more likely to hit that problem we need to persist lots of tiny document IDs, to exaggerate the “error” in the old BSON length estimation logic.

internal/verifier/recheck_test.go
FGasper: TBH I don’t know. I didn’t think it prudent to hold up merging this, though, to solve that.

PR #105, opened 2025-04-16
REP-5963 Migrate recheck queue to per-generation collections

FGasper:
Previously we stored the entire recheck queue in a single collection. This meant that, when we finished a generation, we had to “surgically” delete all rechecks for the old generation.

This changeset alters the metadata so that, rather than a storing rechecks in a single collection, we store them in generation-specific collections like `recheckQueue_gen0`, `recheckQueue_gen1`, etc. That way, when we want to clean out a prior generation’s recheck queue, we just drop a collection, which is much faster than deleting 100s of millions of individual recheck documents.


tdq45gj (approved):
LGTM!

autarch (approved):
LGTM!

PR #106, opened 2025-04-17
Suppress min & max bounds from logs; clean up logging.

FGasper:
This contains 2 commits. The first removes literal min & max values, which are potentially-sensitive customer data, from logs. The second alters the logs in partitions.go to use structured logging rather than creating long messages with the data points. (The actual information is unchanged; it’ll just be more consistent with other messages added recently.)

The commits will be preserved, rather than being squashed, on merge.


autarch (approved):
LGTM!

PR #107, opened 2025-04-17
REP-5971 Reduce $sample cardinality

tdq45gj:
This commit handles `SampleTooManyDuplicates` by reducing the number of partitions to 1/10 of the original value. It also reduces the number of documents to sample per partition from 10 to 3.

### DIFF
--- a/internal/partitions/partitions.go
+++ b/internal/partitions/partitions.go
@@ -11,6 +11,7 @@ import (
 	"github.com/10gen/migration-verifier/internal/types"
 	"github.com/10gen/migration-verifier/internal/util"
 	"github.com/10gen/migration-verifier/internal/uuidutil"
+	"github.com/10gen/migration-verifier/mmongo"
 	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson"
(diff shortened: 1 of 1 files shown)


internal/partitions/partitions.go
```diff
@@ -613,7 +636,7 @@ func getMidIDBounds(
 
 	// Get a cursor for the $sample and $bucketAuto aggregation.
 	var midIDBounds []any
-	agRetryer := retry.New().WithErrorCodes(util.SampleTooManyDuplicatesErrCode)
+	agRetryer := retry.New()
```
FGasper, line 639: Should we not still retry on this error code? That way we’ll only change the sampling after having repeated the request.

FGasper (approved):
I won’t be around to approve this until Tuesday, so use your judgement on the review point I made.

Thank you!

internal/partitions/partitions.go
tdq45gj: It seems unnecessary to retry with same parameters. Especially if the workload is large, each retry attempt would just have the same high probability of failing again.

PR #109, opened 2025-04-23
REP-5996 Allow users to reduce the max size of recheck tasks.

FGasper:
This allows users to set a custom maximum size for recheck tasks. The verifier observes this setting in 2 ways:
1. When creating recheck tasks, it will move onto the next task if the list of IDs has reached the maximum size.
2. When running recheck tasks, it splits the list of IDs into groups and runs each group separately (i.e., in series). This is so that, if a verification finishes generation 0 then hits problems in generation 1, the user can adjust the recheck size and have it take effect without restarting verification.

The reason for this change is that, as seen in HELP-73941, if the verifier sends many large queries to the server at once, it can deplete the server’s available memory, which can cause OOMs and elections. By throttling this new setting, customers can (hopefully) mitigate that problem.

This also reduces the default recheck size from 12 MiB to 8 MiB.

### DIFF
--- a/internal/util/bson.go
+++ b/internal/util/bson.go
@@ -1,8 +1,11 @@
 package util
 
 import (
+	"slices"
 	"strconv"
 
+	"github.com/10gen/migration-verifier/mbson"
+	"github.com/pkg/errors"
 	"go.mongodb.org/mongo-driver/bson"
 )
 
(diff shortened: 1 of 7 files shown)


tdq45gj (approved):
LGTM

internal/util/bson.go
```diff
@@ -35,3 +38,37 @@ func (b *BSONArraySizer) Len() int {
 	// 4 bytes for BSON length, the elems, then a final NUL.
 	return 5 + b.elemsSize
 }
+
+// SplitArrayByBSONMaxSize takes an array of arbitrary Go values and
+// groups them so that, when built into a BSON array, each group exceeds
+// maxSize by the smallest length possible.
```
mmcclimon, line 44: (this feels like a weird definition of "max size", in that every element will exceed it!)

mmcclimon (approved):
LGTM % two non-blocking questions/comments, thanks.

internal/util/bson.go
FGasper: Well, every element _but_, we assume, the last.

I wasn’t quite sure what to call it … `softMaxSize`?

internal/verifier/migration_verifier.go
```diff
@@ -598,10 +608,54 @@ func (verifier *Verifier) ProcessVerifyTask(ctx context.Context, workerNum int,
...
+					Int("workerNum", workerNum).
+					Any("task", task.PrimaryKey).
+					Str("namespace", task.QueryFilter.Namespace).
+					Int("subsetDocs", len(ids)).
+					Int("totalTaskDocs", len(task.Ids)).
+					Msg("Comparing subset of recheck task’s documents.")
+			}
+
+			miniTask := *task
+			miniTask.Ids = ids
```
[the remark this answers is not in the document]
FGasper: I added a clone.Clone() around this to prevent that.

PR #111, opened 2025-05-01
REP-5806 Parse document filter as extended JSON

FGasper:
Previously we parsed the document filter as plain JSON, which made it impossible to filter on ObjectID or other BSON-specific types.

This changeset fixes that by parsing the document filter as MongoDB’s extended JSON.

I’m going to merge the commits as-is rather than squashing. Note that there are two unrelated UX improvements here, as well as a fix for a race condition in an unrelated test, in discrete commits.

### DIFF
--- a/internal/verifier/migration_verifier_test.go
+++ b/internal/verifier/migration_verifier_test.go
@@ -1642,14 +1642,36 @@ func (suite *IntegrationTestSuite) TestVerifierWithFilter() {
 	<-checkDoneChan
 }
 
-func (suite *IntegrationTestSuite) waitForRecheckDocs(verifier *Verifier) {
+func (suite *IntegrationTestSuite) awaitEnqueueOfRechecks(verifier *Verifier, minDocs int) {
+	var lastNonzeroRechecksCount int
+
 	suite.Eventually(func() bool {
 		cursor, err := verifier.getRecheckQueueCollection(verifier.generation).
 			Find(suite.Context(), bson.D{})
-		var docs []bson.D
+		var rechecks []bson.D
 		suite.Require().NoError(err)
-		suite.Require().NoError(cursor.All(suite.Context(), &docs))
-		return len(docs) > 0
+		suite.Require().NoError(cursor.All(suite.Context(), &rechecks))
+
+		if len(rechecks) >= minDocs {
+			return true
+		}
+
+		if len(rechecks) > 0 {
+			// Note any progress toward our minimum rechecks count.
+			if lastNonzeroRechecksCount != len(rechecks) {
+				suite.T().Logf(
+					"%d recheck(s) are enqueued, but we want %d.",
+					len(rechecks),
+					minDocs,
+				)
+
+				lastNonzeroRechecksCount = len(rechecks)
+			}
+
+			return false
+		}
+
+		return false
 	}, 1*time.Minute, 100*time.Millisecond)
 }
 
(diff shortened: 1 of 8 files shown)


internal/verifier/webserver/binding_extjson.go
```diff
@@ -0,0 +1,48 @@
...
+
+func decodeExtJSON(rdr io.Reader, obj any) error {
+	fullDoc, err := io.ReadAll(rdr)
+	if err != nil {
+		return fmt.Errorf("reading request: %w", err)
+	}
+
+	err = bson.UnmarshalExtJSON(fullDoc, false, obj)
+
+	if err != nil {
```
ajayvijayakumar123, line 43: I see that in mongosync we [have](https://github.com/10gen/mongosync/blob/ba564edfc848eae72e8df997594758a6a302e5de/internal/webserver/extjson.go#L43) `binding.Validator.ValidateStruct(obj)`. Do we also need that here?

internal/verifier/migration_verifier_test.go
```diff
@@ -1592,14 +1642,36 @@ func (suite *IntegrationTestSuite) TestVerifierWithFilter() {
...
+
 	suite.Eventually(func() bool {
 		cursor, err := verifier.getRecheckQueueCollection(verifier.generation).
 			Find(suite.Context(), bson.D{})
-		var docs []bson.D
+		var rechecks []bson.D
 		suite.Require().NoError(err)
-		suite.Require().NoError(cursor.All(suite.Context(), &docs))
-		return len(docs) > 0
+		suite.Require().NoError(cursor.All(suite.Context(), &rechecks))
```
ajayvijayakumar123, line 1653: Why do we need this change?

ajayvijayakumar123 (changes requested):
Looks good, had some questions!

internal/verifier/migration_verifier_test.go
FGasper: This is to fix a race condition that isn’t really related to the rest of the changeset but happened to appear in testing.

internal/verifier/webserver/binding_extjson.go
FGasper: I should probably just copy-paste the one from mongosync … which, heh, I seem to have written myself. (This _seemed_ awfully familiar …)

internal/verifier/webserver/binding_extjson.go
```diff
@@ -0,0 +1,47 @@
...
+var _ binding.Binding = extJSONBindingType{}
+var _ binding.BindingBody = extJSONBindingType{}
+
+func (extJSONBindingType) Name() string {
+	return "extJSON"
+}
+
+func (ejb extJSONBindingType) Bind(req *http.Request, obj any) error {
+	if req == nil || req.Body == nil {
+		return errors.New("invalid request")
```
tdq45gj, line 26: Should `req.Body == nil` be an error or should it return an empty struct?

tdq45gj (approved):
LGTM % one question about a very trivial case

internal/verifier/webserver/binding_extjson.go
FGasper: I don’t know, actually. I just copied this from mongosync—which is code I wrote sometime back, probably just mimicking gin’s own bindings.

I’m going to leave as-is since I agree it’s a pretty trivial case that’s unlikely to matter.

PR #112, opened 2025-05-06
Fix ConvertToRawValue, which is necessary if _id is BSON null.

FGasper:
This fixes a small issue that arose in mongosync testing: if a document’s `_id` is BSON null—which can happen!—then the logic to chunk a slice of document `_id`s failed because apparently bson.MarshalValue() fails if given Go’s nil.


mmcclimon (approved):
Oh sure, fine by me. It would nice if the commit message also _described_ the issue that turned up in testing. (Like, why did we just notice this now? Was it a new test, or did something unexpectedly get nil, or what?) But the code is just fine with me, thanks!

PR #113, opened 2025-05-07
REP-5972 Version metadata to prevent breakage across verifier builds.

FGasper:
This adds a metadata version to the persisted generation and makes migration-verifier fail if it starts up and finds a persisted metadata version that mismatches its own.


internal/verifier/generation.go
```diff
@@ -16,6 +16,22 @@ const (
 	generationFieldName = "generation"
 )
 
+type generationDoc struct {
+	Generation      int `bson:"generation"`
+	MetadataVersion int
```
autarch, line 21: Does this not need a `bson` tag? If it doesn't, then why does `Generation` need it?

autarch (approved):
LGTM % one question.

internal/verifier/generation.go
FGasper: Neither of these probably need the tag. `Generation` has it, I think, just because it’s always had it.

FGasper: Removed after confirming that it make no difference as written.

PR #116, opened 2025-05-29
Add shell script parameters to support build for different architectures and operating systems

ShiriNmi1520:
## Description
I've modified the shell script to be able to build for different architectures and operating systems.
For fallback options, if the user didn't provide any parameters, it will build for the current environment using `go env` to get essential variables.
## Screenshot
Usage prompt
image

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,6 +1,38 @@
 #!/bin/sh
 
+usage() {
+cat << EOF
+Usage: ./build.sh [-a <386|amd64|arm|arm64>] [-o <linux|darwin|windows>]
+Build migration-verifier for mongosync.
+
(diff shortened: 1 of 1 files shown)


ShiriNmi1520:
> What does this script do that just manually setting GOARCH and GOOS doesn’t? ([example](https://github.com/mongodb-labs/migration-verifier/blob/916bad0ee16a13117028f73338f9d64f9738c4a4/.github/workflows/release.yml#L31-L33))

Hi @FGasper, I checked with the provided link and found the release workflow already built for different os and architectures.
I think this PR is good to close now. Thanks for pointing it out.

PR #118, opened 2025-06-03
REP-6129 Accommodate type bracketing for pre-v5 find queries

FGasper:
Pre-v5 server versions don’t recognize indexes in $expr-using partition queries. As a result, such queries entail a collection scan. To prevent the performance problems that that would create, migration-verifier avoids $expr when querying partitions against these server versions. Thus, migration-verifier’s partition queries have been type-bracketed; i.e., they only return results for types in the query itself.

For example, consider a query for all documents with `_id` between MinKey and ObjectID(...). An $expr query will return any document with an `_id` that sorts between those two values. A non-$expr query, though, will only return documents with an `_id` of one of those 2 types, despite the fact that many types—including strings and all numeric types—sort between MinKey and ObjectID.

As a result of its non-$expr queries, migration-verifier has been quietly skipping documents when querying pre-v5 servers. In migrations to v5+ it’s a minor concern because whatever documents migration-verifier skips on the source will be recorded as “missing” in generation 0 but will “magically” appear in generation 1. (The only problem is when large numbers of “missing” documents happen, which caused REP-6088.)

In migrations to pre-v5 versions, though, migration-verifier will simply skip documents.

This changeset fixes this problem in pre-v5 partition queries by:
- determining the partition bounds’ BSON types
- determining the BSON types that type bracketing will exclude
- adding an $or clause that includes documents with `_id`s of those types

The above example, then, becomes:
```
(_id >= MinKey OR type(_id) > MinKey)
AND
(_id <= ObjectID(...) OR type(_id) < ObjectID)
```

NB: mongosync solves this problem differently. It uses $expr but adds hint, min, and max to the query. This properly constrains the query, but it excludes the partition’s upper bound. To compensate for that, monosync adds a separate, single-value partition to check the partition’s upper bound. This changeset uses the above-described approach instead because it seems simpler and less disruptive.


autarch (approved):
This all looks good to me, except one spot which I think could use a comment.

PR #123, opened 2025-07-29
Fix handling of non-transient errors when processing doc-compare tasks.

FGasper:
Task failures weren’t being reported “up the chain” because of a scoping bug in the task-execution code. This changeset fixes that bug, so now non-transient task errors will crash the verifier.


autarch (approved):
LGTM!

PR #124, opened 2025-08-01
Switch to annotated tags for releases

FGasper:
Annotated tags conveniently facilitate bundling release notes. This changes the build process to require an annotated tag.


mmcclimon (approved):
LGTM, thanks!

PR #126, opened 2025-08-11
Avoid CGO when building.

FGasper:
This prevents problems when running migration-verifier on older OSes like Amazon Linux 2 (that remain in use).


mmcclimon (approved):
I mean, approving because it successfully disables cgo, but I have no idea why we want to do this or what problem it's meant to solve. 🤷‍♂️

PR #127, opened 2025-08-13
REP-6465 Fix handling of dotted shard keys

FGasper:
This fixes Migration Verifier’s logic to assemble a document key given a full document and field names when any of those fields has a dot.

When the server routes a document whose collection is sharded on, e.g., `{foo.bar: 1}`, it does so based on the value of /foo/bar (i.e., the `bar` field of the top-level embedded document named `foo`) in the document. Migration Verifier instead assembled its document key from the value at /foo.bar (i.e., a top-level field named `foo.bar`).

This fixes the bug. Notably, it also avoids the problem, documented in SERVER-109340, where the server’s change stream’s `documentKey` inaccurately reports the values used to route the document.


internal/verifier/compare.go
```diff
@@ -356,21 +355,110 @@ func (verifier *Verifier) compareDocsFromChannels(
...
+) (bson.Raw, error) {
+	var dk bson.D
+	for _, field := range fieldNames {
+
+		// This is how sharding routes documents: it always
+		// splits on the dot and looks deeply into the document.
+		parts := strings.Split(field, ".")
+		val, err := doc.LookupErr(parts...)
+
+		if errors.Is(err, bsoncore.ErrElementNotFound) {
```
mmcclimon, line 443: You probably also want to handle `bsoncore.ErrInvalidDepthTraversalError` here, which is the kind of error you get if you have a shard key like `a.b` and the `a` field in the target document is an array.

internal/verifier/dockey_agg_test.go
```diff
@@ -0,0 +1,195 @@
...
+				OperationType  string
+				FullDocument   bson.D
+				DocumentKey    bson.D
+				ComputedDocKey bson.D
+			}
+
+			require.NoError(bson.Unmarshal(changes.Current, &event))
+
+			if event.OperationType != "insert" {
+				t.Logf("Ignoring irrelevant-seeing event: %v", changes.Current)
```
mmcclimon, line 173:
```suggestion
				t.Logf("Ignoring irrelevant-seeming event: %v", changes.Current)
```

internal/verifier/dockey_agg_test.go
```diff
@@ -0,0 +1,195 @@
...
+				"checking computed doc key for %v against server",
+				curCase.doc,
+			)
+
+			break
+		}
+
+		require.NoError(changes.Err(), "change stream should not fail")
+
+		changes.Close(ctx)
```
mmcclimon, line 193: Typically this is called in a `defer` right after we open successfully.

mmcclimon (approved):
I mean, this seems fine? I'm deeply suspicious of the aggregation pipeline because I do not feel at all qualified to review that, so you might also get a review from someone that does. The tests seem to pass though, but you might want some more cases for weird things (notably, arrays where you might get depth-traversal errors).

PR #128, opened 2025-08-18
REP-6492 Switch to $sampleRate-style partitioning

FGasper:
$sample-based partitioning has proven problematic for some years now because it often creates highly-imbalanced partitions.

This changeset switches partitioning to use $sampleRate instead. Because this entails a full index scan it tends to be slower; we offset that by creating partition tasks immediately as we receive sampled partition boundaries rather than all at once at the end of the aggregation.

Because MongoDB 4.2 lacked $sampleRate (and $rand as well), the legacy partitioning logic remains for use with that server version.

Both legacy & $sampleRate partitioning are made to use `available` read concern and `secondaryPreferred` read preference. These aggregations don’t need consistency, but they benefit substantially from speed & minimizing workload on the primary.

A few simplifications are made here as well. For example, `MongosyncID` is removed from the PartitionKey struct since it’s never actually relevant, and certain parameters to the legacy partitioner are made constant (since they were always used thus). The `util.Divide` function is renamed `util.DivideToF64` to clarify that its return is a float.

### DIFF
--- a/internal/verifier/partition.go
+++ b/internal/verifier/partition.go
@@ -198,14 +198,14 @@ func (verifier *Verifier) createPartitionTasksWithSampleRateRetryable(
 		return nil
 	}
 
-	idealNumPartitions := util.Divide(collBytes, idealPartitionBytes)
+	idealNumPartitions := util.DivideToF64(collBytes, idealPartitionBytes)
 
 	// We only want to go in here when the collection has enough data
 	// to justify partitioning.
 	if idealNumPartitions > 1 {
-		docsPerPartition := util.Divide(docsCount, idealNumPartitions)
+		docsPerPartition := util.DivideToF64(docsCount, idealNumPartitions)
 
-		sampleRate := util.Divide(1, docsPerPartition)
+		sampleRate := util.DivideToF64(1, docsPerPartition)
 
 		if sampleRate > 0 && sampleRate < 1 {
 			pipeline = append(
--- a/internal/reportutils/reportutils.go
+++ b/internal/reportutils/reportutils.go
@@ -95,7 +95,7 @@ func BytesToUnit[T num16Plus](count T, unit DataUnit) string {
 		panic(fmt.Sprintf("Missing unit in unitSize: %s", unit))
 	}
 
-	return FmtReal(util.Divide(count, myUnitSize))
+	return FmtReal(util.DivideToF64(count, myUnitSize))
 }
 
 // FmtReal provides a standard formatting of real numbers, with a consistent
@@ -128,7 +128,7 @@ func roundFloat(val float64, precision uint) float64 {
 }
 
 func fmtQuotient[T, U realNum](dividend T, divisor U) string {
-	return FmtReal(util.Divide(dividend, divisor))
+	return FmtReal(util.DivideToF64(dividend, divisor))
 }
 
 // FmtPercent returns a stringified percentage without a trailing `%`,
--- a/internal/util/math.go
+++ b/internal/util/math.go
@@ -2,7 +2,7 @@ package util
 
 import "github.com/10gen/migration-verifier/internal/types"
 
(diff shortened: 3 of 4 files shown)


internal/verifier/partition.go
```diff
@@ -0,0 +1,260 @@
...
+			CollName: srcColl.Name(),
+		},
+	)
+	if err != nil {
+		return 0, 0, 0, errors.Wrapf(err, "getting %#q’s shard key", srcNs)
+	}
+
+	pipeline := mongo.Pipeline{
+		{{"$project", bson.D{{"_id", 1}}}},
+		{{"$sort", bson.D{{"_id", 1}}}},
```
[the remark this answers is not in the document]
FGasper: Whew! fixed

internal/verifier/migration_verifier_test.go
```diff
@@ -68,6 +68,68 @@ func TestIntegration(t *testing.T) {
...
+			ctx,
+			task,
+		)
+		require.NoError(err, "should insert collection task")
+
+		require.NoError(
+			verifier.ProcessCollectionVerificationTask(ctx, 0, task),
+			"task should succeed",
+		)
+	*/
```
tdq45gj, line 103: Do we still need this in the test?

internal/verifier/partition.go
```diff
@@ -95,8 +95,9 @@ func (verifier *Verifier) createPartitionTasksWithSampleRateRetryable(
 	}
 
 	pipeline := mongo.Pipeline{
-		{{"$project", bson.D{{"_id", 1}}}},
+		// NB: $sort MUST precede $project in order to avoid a blocking sort.
```
tdq45gj, line 98: Could be more specific that it's only a problem for versions before v6.0.

internal/verifier/partition.go
```diff
@@ -225,24 +197,56 @@ func (verifier *Verifier) createPartitionTasksWithSampleRateRetryable(
 		return nil
 	}
 
-	lowerBound := lowerBoundOpt.OrElse(primitive.MinKey{})
+	idealNumPartitions := util.Divide(collBytes, idealPartitionBytes)
+	docsPerPartition := util.Divide(docsCount, idealNumPartitions)
```
tdq45gj, line 201: It seems that it's still possible to get division by 0 error. If `collBytes  < idealPartitionBytes`, `idealNumPartitions` would become 0. Could we limit `idealNumPartitions` and `docsPerPartition` to be larger than 1?
```suggestion
	idealNumPartitions := min(1, util.Divide(collBytes, idealPartitionBytes))
	docsPerPartition := min(1, util.Divide(docsCount, idealNumPartitions))
```

internal/verifier/partition.go
```diff
@@ -0,0 +1,260 @@
...
+		srcColl,
+	)
+	if err != nil {
+		return 0, 0, 0, errors.Wrapf(err, "getting %#q’s size", srcNs)
+	}
+
+	idealNumPartitions := util.Divide(collBytes, idealPartitionBytes)
+
+	docsPerPartition := util.Divide(docsCount, idealNumPartitions)
+	sampleRate := util.Divide(1, docsPerPartition)
```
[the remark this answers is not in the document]
tdq45gj: Oh I think because we're converting to float in `util.Divide`, division by 0 results in `+inf`.

internal/verifier/partition.go
FGasper: done

internal/verifier/migration_verifier_test.go
FGasper: oops, removed

internal/verifier/partition.go
FGasper: util.Divide() returns a float, so that shouldn’t happen, but I’ve tweaked the logic anyhow.

PR #130, opened 2025-08-20
REP-6438 Include currently-running task stats in baseline logs.

FGasper:
Currently the non-debug logs offer no insight into currently-active tasks’ progress. So if tasks take a long time, non-debug logging will show now progress until one of those tasks finishes.

This amends logging thus:
- Document & byte tallies now incorporate in-progress tasks’ document & byte counts.
- “Active document comparison threads” is now printed.
- Each line of the “Namespaces in progress” table shows that namespace’s count of active threads.


tdq45gj (approved):
LGTM

PR #134, opened 2025-09-04
Add a script to download the latest version.

FGasper:
This facilitates use in CI.

### DIFF
--- a/download_latest.sh
+++ b/download_latest.sh
@@ -36,7 +36,7 @@ fi
 
 echo "Downloading $DOWNLOAD_URL …"
 
-curl -L "$DOWNLOAD_URL" > "$filename"
+curl -sSL "$DOWNLOAD_URL" > "$filename"
 
 chmod +x "$filename"
 
--- a/download_latest.sh
+++ b/download_latest.sh
@@ -6,7 +6,7 @@ filename=migration_verifier
 
 RELEASE_URL="https://api.github.com/repos/mongodb-labs/migration-verifier/releases/latest"
 
-OS=$(uname -o | tr '[:upper:]' '[:lower:]')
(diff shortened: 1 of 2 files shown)


download_latest.sh
```diff
@@ -0,0 +1,43 @@
...
+
+OS=$(uname -o | tr '[:upper:]' '[:lower:]')
+ARCH=$(uname -m)
+if [ "$ARCH" = "aarch64" ]; then
+    ARCH=arm64
+fi
+
+echo "Looks like you’re running $OS on $ARCH."
+
+MANIFEST=$(wget -qO- "$RELEASE_URL")
```
autarch, line 17: The instructions tell people to use `curl` to fetch this script, but then it uses `wget` internally. I think it should `curl` since we know they already have that installed.

autarch: But maybe this isn't needed at all.

download_latest.sh
autarch, line 1: I think this is more complex than it needs to be. Once you know the filename you want to download, you can download this: `https://github.com/mongodb-labs/migration-verifier/releases/latest/download/$FILENAME`

There's no need to hit the API and parse the response.

This does mean you need to calculate the filename directly instead of getting it from the API response, but it seems a lot simpler. You can see how I do this for ubi at https://github.com/houseabsolute/ubi/blob/master/bootstrap/bootstrap-ubi.sh

autarch (changes requested):
I think there's a simpler way to do this.

download_latest.sh
FGasper: I’d rather not build in a dependency on the filename format.

OK to keep as-is?

autarch (approved):
LGTM!

PR #138, opened 2025-09-17
Make the download script interpret “x86_64” as just “amd64”.

FGasper:


autarch (approved):
LGTM! (cough, `ubi`, cough)

PR #140, opened 2025-09-18
Add time-series support

FGasper:
This changeset adds time-series support. It works with & without namespace filtering.

Because of the verifier’s `_id` dependency, this verifies buckets rather than logical measurements. This implies a requirement that the migration copy time-series collections via buckets as well since a logical replication would not preserve artifacts like bucket `_id` or even the grouping of measurements.

Because of the bucket-level verification, details in mismatch reports are not very useful for time-series because they reference bucket-level fields that the logical API doesn’t expose.

### DIFF
--- a/internal/verifier/migration_verifier.go
+++ b/internal/verifier/migration_verifier.go
@@ -1120,7 +1120,6 @@ func (verifier *Verifier) verifyMetadataAndPartitionCollection(
 		}
 		task.Status = verificationTaskMetadataMismatch
 	}
-
 	if !verifyData {
 		// If the metadata mismatched and we're not checking the actual data, that's a complete failure.
 		if task.Status == verificationTaskMetadataMismatch {
--- a/internal/verifier/migration_verifier.go
+++ b/internal/verifier/migration_verifier.go
@@ -827,11 +827,7 @@ func (verifier *Verifier) compareCollectionSpecifications(
 	switch srcSpec.Type {
 	case "collection":
 		canCompareData = true
-	case "view":
-	case "timeseries":
-		if !verifier.verifyAll {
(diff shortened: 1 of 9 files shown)


internal/verifier/migration_verifier.go
```diff
@@ -822,8 +822,20 @@ func (verifier *Verifier) compareCollectionSpecifications(
...
 	}
 
-	// Don't compare view data; they have no data of their own.
-	canCompareData := srcSpec.Type != "view"
+	var canCompareData bool
+
+	switch srcSpec.Type {
+	case "collection":
+		canCompareData = true
+	case "view":
```
tdq45gj, line 830: nit: would prefer `canCompareData` being set to false explicitly.

internal/verifier/timeseries_test.go
```diff
@@ -0,0 +1,353 @@
...
+	"github.com/10gen/migration-verifier/mslices"
+	"github.com/samber/lo"
+	"github.com/stretchr/testify/require"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+// TestTimeSeries_BucketsOnly confirms the verifier’s time-series coverage
+// when only the buckets exist. This is important when verifying shard-to-shard.
```
tdq45gj, line 16: > This is important when verifying shard-to-shard.

Do we test shard-to-shard? Just curious.

tdq45gj (approved):
LGTM

internal/verifier/timeseries_test.go
FGasper: Shard-to-shard verification has been tested, yeah. Nothing is automated, though.

PR #141, opened 2025-09-19
Re-add 4.2 in CI.

FGasper:
`m`’s maintainer has fixed the issue that caused 4.2 downloads to fail, so this changeset re-enables 4.2 testing in CI.


autarch (approved):
LGTM!

PR #290, opened 2026-05-13
Replace internal Option type with migration-tools’s version

FGasper:
migration-tools’s Option enjoys better testing, particularly to prevent unneeded heap allocations.

This also adds a few preexisting allocation tests that had yet to be committed.


tdq45gj (approved):
LGTM

PR #291, opened 2026-05-13
REP-7312 Add limited DDL support

FGasper:
This adds limited support for DDL change events:
- `drop` and `rename` remain unsupported.
- Change streams only (not available with oplog tailing).


FGasper:

PR #292, opened 2026-05-13
Fix grouping bug when multiple rechecks occur for the same document.

FGasper:
Previously if enqueued rechecks for the same document straddled a recheck-group boundary, GenerateRecheckTasks would split the dupe rechecks across tasks. Now those rechecks are in the same task.

Example: Assume 9,999 rechecks are ready to be converted to a task. Assume the next 2 rechecks are for the same document. Verifier _should_ deduplicate those rechecks into the same recheck task. Instead it was splitting them across tasks.

This changeset also fixes an off-by-one error in that part of the code that made the effective per-recheck-task document limit 10,001 rather than 10,000.

An additional assertion is added, with an errantly omitted change reader name added.

This should not affect significant user-visible behavior. It’s still good to clear up, though, given the high complexity of this part of the code.

### DIFF
--- a/internal/verifier/recheck.go
+++ b/internal/verifier/recheck.go
@@ -102,6 +102,10 @@ func (verifier *Verifier) insertRecheckDocs(
 		panic("Expect either mismatch times or optimes, not both/neither.")
 	}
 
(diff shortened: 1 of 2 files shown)


tdq45gj (approved):
LGTM

PR #294, opened 2026-05-14
REP-7312 Support most DDL events

FGasper:
This adds support for most DDL events. 

Caveats:
- change streams only support DDL from v6+
- `drop` and `rename` remain unsupported


FGasper:

PR #302, opened 2026-06-04
REP-7412 Add retry to GetCollectionSpecIfExists

FGasper:
This prevents a crash if the fetch of a collection specification fails due to a transient error.

A small refactor is needed to prevent an import cycle.


tdq45gj (approved):
LGTM

PR #314, opened 2026-08-07
Fix direct-connect when the connection string has an argument.

FGasper:
Previously connection strings like `mongodb://example.com?authSource=admin` would have a mangled `directConnection` argument postfixed.

This fixes that and adds a preliminary ping to detect connection problems up-front.

It also redacts the connection string from existing connection errors and deduplicates some of the logic to connect to the clusters.

### DIFF
--- a/internal/verifier/migration_verifier.go
+++ b/internal/verifier/migration_verifier.go
@@ -383,6 +383,10 @@ func (verifier *Verifier) SetMetaURI(ctx context.Context, uri string) error {
 		return err
 	}
 
+	if err := verifier.metaClient.Ping(ctx, readpref.Primary()); err != nil {
+		return errors.Wrapf(err, "connect to metadata server (%s)", uri)
+	}
+
 	verifier.logger.Debug().
 		Msg("Reading metadata’s cluster info.")
 
--- a/internal/verifier/migration_verifier.go
+++ b/internal/verifier/migration_verifier.go
@@ -273,7 +273,11 @@ func (verifier *Verifier) SetDDLHandling(mode DDLHandling) {
 	verifier.ddlHandling = mode
 }
 
-func (verifier *Verifier) getClientOpts(uri string) *options.ClientOptions {
+func (verifier *Verifier) getClient(
+	ctx context.Context,
+	uri string,
+	rp *readpref.ReadPref,
+) (*mongo.Client, *options.ClientOptions, error) {
 	appName := buildvar.GetClientAppName()
 	opts := &options.ClientOptions{
 		AppName: &appName,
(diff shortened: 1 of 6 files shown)


internal/verifier/migration_verifier.go
```diff
@@ -285,7 +289,20 @@ func (verifier *Verifier) getClientOpts(uri string) *options.ClientOptions {
...
+	}
+
+	client, err := mongo.Connect(opts)
+	if err != nil {
+		return nil, nil, fmt.Errorf("create connection: %w", err)
+	}
+
+	if err := client.Ping(ctx, readpref.Primary()); err != nil {
+		return nil, nil, fmt.Errorf("ping: %w", err)
+	}
```
tdq45gj: Should the ping use the readPref that's been set to the client?

internal/verifier/uri.go
```diff
@@ -62,13 +59,13 @@ func isVersionSupported(version []int) bool {
 }
 
 func (verifier *Verifier) SetDstURI(ctx context.Context, uri string) error {
-	opts := verifier.getClientOpts(uri)
-	var err error
-	verifier.dstClient, err = mongo.Connect(opts)
+	client, clientOpts, err := verifier.getClient(ctx, uri, verifier.readPreference)
 	if err != nil {
-		return errors.Wrapf(err, "failed to connect to destination %#q", uri)
+		return errors.Wrapf(err, "connect to destination")
```
tdq45gj, line 65: Should we still include a redacted URI?

tdq45gj: Seems that we're still logging the raw URI in `checkURIAgainstServerVersion`.

tdq45gj (changes requested):
The changes looks good to me. But have we double checked that we're not logging raw URI anywhere else in the MV?

internal/verifier/migration_verifier.go
FGasper: Maybe? It didn’t seem to matter much. But we might as well, probably.

internal/verifier/uri.go
FGasper: I don’t know if MV has logic to do that redaction. (That would make a good candidate for moving to migration-tools …)

tdq45gj: Moving the redaction to migration-tools makes sense to me. Should we remove logging the raw URI in `checkURIAgainstServerVersion` for now?

FGasper: Is there still a log of the uri? I removed the one in the panic …

tdq45gj (approved):
LGTM, thanks!

